PR #4, opened 2020-02-09
Json

rrousselGit:


gadfly361:
@rrousselGit Wow that was fast! Happy to see json_serializable included with this generator. 

Will @JsonKey be usable with freezed?

rrousselGit:
For now, no.
This is the initial support step.

But we can think of:

```dart
abstract class Foo with _$Foo {
 factory  Foo(@JsonKey('my_var') String myVar) = _Foo;
}
```

PR #45, opened 2020-02-14
Allow static getters

### LINKED ISSUE #44: static getter not working since 0.4.0
It seems that static getters on `@freezed` classes are broken since 0.4.0.

The following file:

test.dart:
```dart
import 'package:flutter/foundation.dart';

part 'test.freezed.dart';
part 'test.g.dart';

@freezed
abstract class Test with _$Test{

  const factory Test() = _Test;

  static int get value => 0;
}
```

works fine on 0.3.0, but on 0.4.0+ I get this error:

```
[SEVERE] freezed:freezed on lib/test.dart:
Error running FreezedGenerator
@freezed cannot be used on classes with unimplemented getters
package:freezed_test/test.dart:9:16
  ╷
9 │ abstract class Test with _$Test{

```

I don't see a reason why static getters would not be supported, I think that's a bug.

static methods still work.

knaeckeKami:
Fixes #44 .

Allow static getters on `@freezed` classes.

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -120,7 +120,7 @@ class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
       );
     }
 
-    if (element.fields.isNotEmpty) {
+    if (element.fields.where((field) => !field.isStatic).isNotEmpty) {
       throw InvalidGenerationSourceError(
         '@freezed cannot be used on classes with unimplemented getters',
         element: rawElement,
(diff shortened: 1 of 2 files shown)


rrousselGit:
Hello, thanks!

Could you write some tests for it?
Just adding a valid scenario on one of the files inside the `test/integration` folder should be enough

rrousselGit:
we'll want to test static methods/getters/variables

knaeckeKami:
Of course, will do!

rrousselGit:
Looks good to me, thanks!

When the CI pass I'll merge it 😄

knaeckeKami:
Great :)

OT:
Ugh, does Github really also link people in commit messages containing an "@"? This will really cause many notifications...

PR #71, opened 2020-02-25
Do not add const modifier if user already explicitly added const modifier in @Default() annotations

knaeckeKami:
Sorry for bringing this topic up again, but there's another issue with const modifiers - the current implementation does not check if the user already added the const modifier explicitly, it always adds it, no matter if it's already there or not.

This fixes this (hopefully). I'm not really happy with this code, seems fragile, feel free to improve upon it if you know a better way.


rrousselGit:
Nice, thanks again
I agree that the code is not ideal. Sadly we can't do otherwise

PR #88, opened 2020-03-04
Add information on @JsonSerializable annotation

orestesgaolin:
It wasn't clear how to set JsonSerializable annotation so I added it to Readme.


rrousselGit:
> By default `json_serializable` won't generate calls to `toJson()` on nested classes, so you need to pass `explicit_to_json` flag through `@JsonSerializable` annotation. 

I don't think it makes sense to mention this.
It largely depends on your setup whether you need it or not (otherwise it would always be enabled) and it's the job of `json_serializable` to explain about that flag, not `freezed`

Other than that, looks good to me 😄

orestesgaolin:
Thanks, removed the remark

rrousselGit:
Perfect thanks!

Issue #130, opened 2020-03-30
Remove @/required annotations on generated `when`


rrousselGit:
```dart
@freezed
abstract class ConversationState with _$ConversationState {
  const factory ConversationState.loadInProgress() = ConversationLoadInProgress;
  const factory ConversationState.loadFailure() = ConversationLoadFailure;
  const factory ConversationState.loadSuccess({
    @required List<Conversation> conversations,
  }) = ConversationLoadSuccess;
}
```

generates

```dart
@optionalTypeArgs
  Result when<Result extends Object>({
    @required Result loadInProgress(),
    @required Result loadFailure(),
    @required Result loadSuccess(@required List<Conversation> conversations),
  });
```

Where `loadSuccess.conversations` shouldn't be `@required`

mhmdanas:
Isn't this fixed by #139?

PR #139, opened 2020-04-12
do not create @required annotation for positional arguments #130

hpoul:
Hi,
I have a possible fix for #130 - this creates:

for:
```
@freezed
abstract class RequiredParams with _$RequiredParams {
  const factory RequiredParams({@required String a}) = RequiredParams0;
  const factory RequiredParams.second({@required String a}) = RequiredParams1;
  const factory RequiredParams.third({@required String b}) = RequiredParams2;
}
```

now:
```
  @override
  @optionalTypeArgs
  Result when<Result extends Object>(
    Result $default(String a), {
    @required Result second(String a),
    @required Result third(String b),
  }) {
    assert($default != null);
    assert(second != null);
    assert(third != null);
    return second(a);
  }
```

instead of:
```
  @override
  @optionalTypeArgs
  Result when<Result extends Object>(
    Result $default(@required String a), {
    @required Result second(@required String a),
    @required Result third(@required String b),
  }) {
    assert($default != null);
    assert(second != null);
    assert(third != null);
    return second(a);
  }
```

I have simply passed an option into the `toString` method of whether `isRequired` should be appended. I guess another option would be to implement it similarly to `showDefaultValue` as a property for the Parameter - but i did not quite understand what the `asExpanded` means - which seems to be the method which modifies the `showDefaultValue` property. And I felt it a bit cleaner to make this as a "render" argument.. because the Parameter is still required, it's just rendered as position argument instead of and named argument?

### DIFF
--- a/packages/freezed/lib/src/templates/parameter_template.dart
+++ b/packages/freezed/lib/src/templates/parameter_template.dart
@@ -157,11 +157,7 @@ class ParametersTemplate {
 
   @override
   String toString() {
-    final buffer = StringBuffer()
-      ..writeAll(
-          requiredPositionalParameters
-              .map<String>((p) => p.toString(skipRequired: true)),
-          ', ');
+    final buffer = StringBuffer()..writeAll(requiredPositionalParameters, ', ');
 
     if (buffer.isNotEmpty &&
         (optionalPositionalParameters.isNotEmpty ||
@@ -239,13 +235,32 @@ class Parameter {
   final List<String> decorators;
   final bool showDefaultValue;
 
+  Parameter copyWith({
+    String type,
+    String name,
+    String defaultValueSource,
+    bool isRequired,
+    bool nullable,
+    List<String> decorators,
+    bool showDefaultValue,
+  }) =>
+      Parameter(
+        type: type ?? this.type,
+        name: name ?? this.name,
+        defaultValueSource: defaultValueSource ?? this.defaultValueSource,
+        isRequired: isRequired ?? this.isRequired,
+        nullable: nullable ?? this.nullable,
+        decorators: decorators ?? this.decorators,
+        showDefaultValue: showDefaultValue ?? this.showDefaultValue,
+      );
+
   @override
-  String toString({bool skipRequired = false}) {
+  String toString() {
     var res = '${decorators.join()}  ${type ?? 'dynamic'} $name';
     if (showDefaultValue && defaultValueSource != null) {
       res = '$res = $defaultValueSource';
     }
-    return isRequired && !skipRequired ? '@required $res' : res;
+    return isRequired ? '@required $res' : res;
   }
 }
 
@@ -268,12 +283,12 @@ class LocalParameter extends Parameter {
         );
 
   @override
-  String toString({bool skipRequired = false}) {
+  String toString() {
     var res = '${decorators.join()} this.$name';
     if (showDefaultValue && defaultValueSource != null) {
       res = '$res = $defaultValueSource';
     }
-    return isRequired && !skipRequired ? '@required $res' : res;
+    return isRequired ? '@required $res' : res;
   }
 }
 
@@ -299,8 +314,8 @@ class CallbackParameter extends Parameter {
   final ParametersTemplate parameters;
 
   @override
-  String toString({bool skipRequired = false}) {
+  String toString() {
     var res = '${decorators.join()} $type $name($parameters)';
-    return isRequired && !skipRequired ? '@required $res' : res;
+    return isRequired ? '@required $res' : res;
   }
 }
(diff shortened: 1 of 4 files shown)


packages/freezed/lib/src/templates/parameter_template.dart
```diff
@@ -157,7 +157,11 @@ class ParametersTemplate {
 
   @override
   String toString() {
-    final buffer = StringBuffer()..writeAll(requiredPositionalParameters, ', ');
+    final buffer = StringBuffer()
+      ..writeAll(
+          requiredPositionalParameters
+              .map<String>((p) => p.toString(skipRequired: true)),
```
rrousselGit, line 163: Instead of `p.toString(skipRequired: true)`, I think it would be more logical to have `p.copyWith(isRequired: false)`

packages/freezed/test/integration/multiple_constructors.dart
```diff
@@ -38,3 +38,10 @@ abstract class Recursive with _$Recursive {
   factory Recursive() = RecursiveImpl;
   factory Recursive.next(RecursiveImpl value) = _RecursiveNext;
 }
+
+@freezed
+abstract class RequiredParams with _$RequiredParams {
+  const factory RequiredParams({@required String a}) = RequiredParams0;
+  const factory RequiredParams.second({@required String a}) = RequiredParams1;
+  const factory RequiredParams.third({@required String b}) = RequiredParams2;
```
rrousselGit, line 46: Is there any purpose to the `third` constructor?

packages/freezed/test/map_test.dart
```diff
@@ -298,5 +301,39 @@ void main() {
...
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'multiple_constructors.dart';
+
+void main() {
+  final value = RequiredParams(a: 'a');
+
+  value.when((a) => a, second: (b) => b, third: (c) => c);
+}
+
+void redundant(@required String test) {
```
rrousselGit, line 318: That function seems unused. What is its purpose?

packages/freezed/lib/src/templates/parameter_template.dart
hpoul: Thanks for taking a look - I was a bit hesitant, since it still is required, it's just positional instead of named.. do you still prefer `isRequired` vs. like `showIsRequired`?  It's very similar to the `showDefaultValue` flag? 🤷‍♂️️ - and is t he `asExpanded` method meant as a way to `copyWith` and can it be reused, or better implement a `copyWith` on `Parameter` and use this method in `asExpanded`, seems like it does not much more than copying the `Parameter` objects.

https://github.com/rrousselGit/freezed/blob/f6e1890453d638617e8305da9471222c90c03eda/packages/freezed/lib/src/templates/parameter_template.dart#L134-L148

packages/freezed/test/integration/multiple_constructors.dart
hpoul: No :-) will remove

packages/freezed/test/map_test.dart
hpoul: It's actually just a sanity check to make sure `@required` flag is marked as redundant in that case..

https://github.com/rrousselGit/freezed/blob/79e00b1380103e5ba956d59c7b932a2ca7efc880/packages/freezed/test/map_test.dart#L325-L330

So it will fail if the linter rules change for whatever reason, otherwise it wouldn't be obvious that the test was useless..

packages/freezed/lib/src/templates/parameter_template.dart
rrousselGit: The asExpanded converts LocalParameter into Parameter, so it switches from this.property to Type property.

If you have a better name in mind, feel free to rename it! I don't quite like the name either

rrousselGit: About a copy vs showIsRequired, I don't really mind either of them.
Although it would be simple to do the copy

hpoul: I've now implemented a Parameter.copyWith method and removed my `toString` changes. Maybe you can take another look. 

(fwiw, i think i'm now grasping the `asExpanded` method, i didn't understand that the source parameters were of type `LocalParameter` and not `Parameter`, since it's working with a `List<Parameter>` - tbh i think i'd change this to just being a flag like `isLocalAssigned` so it wouldn't be necessary to have 3 versions of the `copyWith` method (which is currently separated into the Parameter.copyWith, `asExpanded` and `asThis` - but maybe i'm still missing something)

rrousselGit: The ideal scenario would be to be able to run Freezed on its own sources. But don't worry about that 😀

PR #144, opened 2020-04-13
Fix == returning false for classes with other prop

### LINKED ISSUE #138: Member named "other" causes structural equality to fail
As the title says, a class with a member named `other` does not make objects properly equatable in version **0.10.5**.

Test case is as follows:
```
@freezed
abstract class A implements _$A {
  const A._();
  const factory A({
    @Default('') String other,
  }) = _A;
}

...

print(A() == A());  // Prints false
```

Note that renaming `other` to `others` (or any other valid variable name) prints out `true`.

mhmdanas:
Fixes #138.

By the way, one of the tests is failing mysteriously on my machine, specifically this one:
```
00:46 +127 -1: test/multiple_constructors_test.dart: tear off [E]
  Expected: Type:<(String, {int b}) => NoCommonParam0>
    Actual: Type:<(String, {int b}) => NoCommonParam0>

  package:test_api                           expect
  test/multiple_constructors_test.dart 41:5  main.<fn>
```
This test fails even on `master`. Is it only happening with me, or are others also having this issue?

### DIFF
--- a/packages/freezed/test/equal_test.dart
+++ b/packages/freezed/test/equal_test.dart
@@ -139,4 +139,8 @@ void main() {
       GenericIterable([1]),
     );
   });
+  test('ObjectWithOtherProperty', () {
+    expect(ObjectWithOtherProperty(1), ObjectWithOtherProperty(1));
+    expect(ObjectWithOtherProperty(1), isNot(ObjectWithOtherProperty(2)));
+  });
 }
--- a/packages/freezed/test/equal_test.dart
+++ b/packages/freezed/test/equal_test.dart
@@ -140,7 +140,12 @@ void main() {
     );
   });
   test('ObjectWithOtherProperty', () {
-    expect(ObjectWithOtherProperty(1), ObjectWithOtherProperty(1));
-    expect(ObjectWithOtherProperty(1), isNot(ObjectWithOtherProperty(2)));
+    expect(ObjectWithOtherProperty(const [2, 3]),
+        ObjectWithOtherProperty(const [2, 3]));
+    expect(ObjectWithOtherProperty(const [2, 3]),
+        isNot(ObjectWithOtherProperty(const [2, 4])));
+    expect(ObjectWithOtherProperty([1, 2]), ObjectWithOtherProperty([1, 2]));
+    expect(
+        ObjectWithOtherProperty([1, 2]), isNot(ObjectWithOtherProperty([1])));
   });
 }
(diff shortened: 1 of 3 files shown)


packages/freezed/test/equal_test.dart
```diff
@@ -139,4 +139,8 @@ void main() {
       GenericIterable([1]),
     );
   });
+  test('ObjectWithOtherProperty', () {
+    expect(ObjectWithOtherProperty(1), ObjectWithOtherProperty(1));
+    expect(ObjectWithOtherProperty(1), isNot(ObjectWithOtherProperty(2)));
```
rrousselGit, line 144: Could you add a test for the "DeepCollectionEquality" part too?

mhmdanas: Do you mean adding another class in which `other` is a `List` instead of an `int` and then adding a test for it?

rrousselGit: Something where the properties are not `identical`

The current test tests only the first part of the changes: `identical(other.${p.name}, this.${p.name})`

mhmdanas: Got it, I'll do it right away!

mhmdanas: What do you think about making `ObjectWithOtherProperty` take dynamic? I think that will make it easier as the class name will get too long without it.

rrousselGit: What's the relationship with `dynamic` and the class name?

mhmdanas: Nothing, I just think that using separate classes, one with `int other` and another with `List<int> other`, will make the class names to verbose. Something like this: `ObjectWithListOtherPropert`.

rrousselGit: There's no need for two classes. A single class that takes a `List` is enough

mhmdanas: But how will we test for `identical(other.${p.name}, this.${p.name}`? By making the lists `const` sometimes and sometimes not, perhaps?

rrousselGit: Yes

mhmdanas: OK, commit coming in a minutes!

rmoralesd:
Hello. We are facing a similar issue when running some tests.

ERROR: Expected: Right<RepositoryFailure, List<SubSpaces>>:<Right([SubSpaces(spaceId: Value(Right(rqIHOuVDXdwRid6plQCY)), name: test1, location: test1)])>
  Actual: Right<RepositoryFailure, List<SubSpaces>>:<Right([SubSpaces(spaceId: Value(Right(rqIHOuVDXdwRid6plQCY)), name: test1, location: test1)])>

Any idea how to solve this?. 
thanks.

mhmdanas:
@rmoralesd you should open an issue for that. Also, can you provide the class definitions?

PR #169, opened 2020-05-12
Sync @JsonKey options with freezed decorators

mseijas:
Hi @rrousselGit , 

Our team has run into some issues where the freezed annotations that we define sometimes either clash with the default `@JsonKey` settings we get, or we're forced to duplicate some semantics around how we want our models to operate.

For example:
```dart
class User with _$User {
 	factory User({
		@required @JsonKey(required: true, disallowNullValues: true) name,
		@nullable @JsonKey(disallowNullValues: false) phoneNumber,
	}) = _User;
}
```

It would be great if we could look at the `@required` and `@nullable` annotations and make sure that the options set on `@JsonKey` also reflect those settings.

I took a stab at this, would love to know your thoughts!

### DIFF
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -1,9 +1,9 @@
 import 'package:analyzer/dart/element/element.dart';
 import 'package:freezed/src/freezed_generator.dart';
 import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:freezed_annotation/freezed_annotation.dart' as annotations;
 import 'package:meta/meta.dart';
 import 'package:source_gen/source_gen.dart';
-import 'package:freezed_annotation/freezed_annotation.dart' as annotations;
 
 import 'copy_with.dart';
 import 'parameter_template.dart';
@@ -120,9 +120,6 @@ ${copyWith.abstractCopyWithGetter}
   String get _properties {
     return constructor.impliedProperties.map((p) {
       var res = '@override $p';
-      if (p.defaultValueSource != null && !p.hasJsonKey) {
-        res = '@JsonKey(defaultValue: ${p.defaultValueSource}) $res';
-      }
       return res;
     }).join();
   }
@@ -349,7 +346,9 @@ class Property {
   final List<String> decorators;
   final bool nullable;
   final String defaultValueSource;
-  final bool hasJsonKey;
+  final JsonKeyOptions jsonKeyOptions;
+
+  bool get hasJsonKey => jsonKeyOptions != null;
 
   Property({
     @required String type,
@@ -357,7 +356,7 @@ class Property {
     @required this.decorators,
     @required this.nullable,
     @required this.defaultValueSource,
-    @required this.hasJsonKey,
+    this.jsonKeyOptions,
   }) : type = type ?? 'dynamic';
 
   factory Property.fromParameter(ParameterElement element) {
@@ -370,23 +369,88 @@ class Property {
       );
     }
 
+    final decorators = parseDecorators(element.metadata);
+
     return Property(
       name: element.name,
       type: parseTypeSource(element),
-      decorators: parseDecorators(element.metadata),
+      decorators: parseDecorators(element.metadata).toList()
+        ..removeWhere((d) => d.contains('@JsonKey')),
       nullable: element.isNullable,
       defaultValueSource: defaultValue,
-      hasJsonKey: element.hasJsonKey,
+      jsonKeyOptions: JsonKeyOptions.fromDecorators(decorators),
     );
   }
 
   @override
   String toString() {
-    return '${decorators.join()} final $type $name;';
+    var jsonKeyOptions = this.jsonKeyOptions ?? JsonKeyOptions();
+
+    if (nullable) {
+      jsonKeyOptions = jsonKeyOptions.mergeValues({
+        'required': 'false',
+        'disallowNullValue': 'false',
+      });
+    } else {
+      jsonKeyOptions = jsonKeyOptions.mergeValues({
+        'required': 'true',
+        'disallowNullValue': 'true',
+      });
+    }
+
+    if (defaultValueSource != null) {
+      jsonKeyOptions = jsonKeyOptions.mergeValues({
+        'defaultValue': defaultValueSource,
+      });
+    }
+
+    return '${decorators.join()} $jsonKeyOptions final $type $name;';
   }
 
   Getter get getter => Getter(
       name: name, type: type, decorators: decorators, nullable: nullable);
+}
+
+class JsonKeyOptions {
+  final Map<String, String> values;
+
+  JsonKeyOptions([this.values = const {}]);
+
+  static JsonKeyOptions fromDecorators(List<String> decorators) {
+    if (decorators == null || decorators.isEmpty) return null;
+
+    String jsonKeyOptions;
+    for (final d in decorators) {
+      if (!d.startsWith('@JsonKey')) continue;
+      jsonKeyOptions = d.split('@JsonKey(')[1].split(')')[0];
+      break;
+    }
+
+    if (jsonKeyOptions == null) return null;
+
+    var values = <String, String>{};
+
+    for (final option in jsonKeyOptions.split(',')) {
+      final part = option.split(':');
+      final name = part[0].trim();
+      final value = part[1].trim();
+      if (name.isNotEmpty && value.isNotEmpty) {
+        values[name] = value;
+      }
+    }
+
+    return JsonKeyOptions(values);
+  }
+
+  @override
+  String toString() {
+    if (values.isEmpty) return '';
+    return '@JsonKey(${values.entries.map((e) => '${e.key}: ${e.value}').join(',')})';
+  }
+
+  JsonKeyOptions mergeValues(Map<String, String> values) {
+    return JsonKeyOptions({}..addAll(this.values)..addAll(values));
+  }
 }
 
 class Getter {
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -312,7 +312,7 @@ class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
                 nullable: p.nullable,
                 defaultValueSource: p.defaultValueSource,
                 // TODO: support hasJsonKey
-                hasJsonKey: false,
+                // hasJsonKey: false,
               ))
           .toList(),
       needsJsonSerializable: globalData.hasJson &&


packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -370,25 +369,90 @@ class Property {
...
+    }
+
+    return '${decorators.join()} $jsonKeyOptions final $type $name;';
   }
 
   Getter get getter => Getter(
       name: name, type: type, decorators: decorators, nullable: nullable);
 }
 
+class JsonKeyOptions {
```
rrousselGit, line 414: This should be done by the Property class instead

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -370,25 +369,90 @@ class Property {
...
+  final Map<String, String> values;
+
+  JsonKeyOptions([this.values = const {}]);
+
+  static JsonKeyOptions fromDecorators(List<String> decorators) {
+    if (decorators == null || decorators.isEmpty) return null;
+
+    String jsonKeyOptions;
+    for (final d in decorators) {
+      if (!d.startsWith('@JsonKey')) continue;
```
rrousselGit, line 424: There are extensions already for such thing. Search for the extensions on the project, you should find them

This shouldn't be done with a `.split`

rrousselGit:
Could you add tests for this PR?

packages/freezed/lib/src/templates/concrete_template.dart
mseijas: 👍

mseijas:
> Could you add tests for this PR? And fix the CI too

Yeah for sure - just wanted to make sure this seemed like it was going in the right direction to you first. Will add!

mseijas:
@rrousselGit Updated the implementation based on your suggestions, and also added tests. Let me know what you think.

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -370,41 +367,123 @@ class Property {
       );
     }
 
+    final jsonKeyValues = element.jsonKeyValues;
+    if (defaultValue != null &&
+        jsonKeyValues != null &&
+        jsonKeyValues.containsKey('defaultValue')) {
+      throw InvalidGenerationSourceError(
+        '@Default cannot be used in conjunction with @JsonKey(defaultValue:)',
```
rrousselGit, line 375: Why is this needed?

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -370,41 +367,123 @@ class Property {
...
+        todo: 'Remove the defaultValue parameter from @JsonKey and only'
+            'use the @Default decorator instead',
+        element: element,
+      );
+    }
+
+    if (element.hasRequired &&
+        jsonKeyValues != null &&
+        jsonKeyValues.containsKey('required')) {
+      throw InvalidGenerationSourceError(
```
rrousselGit, line 385: Same here, why is this needed?

packages/freezed/test/json_test.dart
```diff
@@ -28,10 +29,38 @@ Future<void> main() async {
       DefaultValue(42),
     );
   });
-  test('@Default does not imply a @JsonKey if one is already specified', () {
+  test('@required implies a @JsonKey', () {
```
rrousselGit, line 32: What is the purpose of this behavior?

Before, it threw a null exception anyway, so I don't see what this adds

packages/freezed/test/json_test.dart
```diff
@@ -28,10 +29,38 @@ Future<void> main() async {
...
+      () => RequiredValue.fromJson(<String, dynamic>{}),
+      throwsA(isA<CheckedFromJsonException>()),
+    );
+
+    expect(
+      () => RequiredValue.fromJson(<String, dynamic>{'value': null}),
+      throwsA(isA<CheckedFromJsonException>()),
+    );
+  });
+  test('@nullable implies a @JsonKey', () {
```
rrousselGit, line 49: I don't understand what this test does, since this is already the current behavior

packages/freezed/lib/src/templates/concrete_template.dart
mseijas: So I added this for safety, to prevent someone writing something like this:
```dart
factory User({@Default(42) @JsonKey(defaultValue: 30) int age})
```
At the call site, it is unclear what behavior will result from this.

packages/freezed/lib/src/templates/concrete_template.dart
mseijas: This is because you could technically write something like this which would also be unclear in terms of what the outcome will be
```dart
factory User({@required @JsonKey(required: false) int age})
```

packages/freezed/test/json_test.dart
mseijas: So before it threw an `AssertionError` -- which is fine in debug builds, but would not get thrown in release builds. This would mean that if deserialization failed in production you would have no way of knowing and the app could end up in a bad state.

What this new behavior does is produce a `CheckedFromJsonException`-- meaning that, since it's not an assertion error, it would still be produced in release builds.

packages/freezed/test/json_test.dart
mseijas: Fair, maybe this is not the way to test this. I just wanted to make sure that if we had the `@nullable` decorator that we would add the corresponding decorator: `@JsonKey(required: false, disallowNullValue: false)`

Do you have any suggestions on how I can test this?

mseijas: For added context, the reason why I think this is important is because someone could write:
```dart
factory User({@nullable @JsonKey(disallowNullValue: true) int age})
```
Resulting again in an indeterministic behavior

rrousselGit:
To be honest, I still don't quite understand what you are trying to do.
Could you explain in further detail the problem?

Maybe what you want is instead to change the default behavior of `@JsonKey` to match `@required` and `@nullable`?

packages/freezed/test/json_test.dart
rrousselGit: The behavior is fixed: `@JsonKey` inside the constructor explicitly takes over every other configuration.

packages/freezed/test/json_test.dart
rrousselGit: That's a fair point, but it's unrelated to deserialisation.
If we want to throw errors on release, this should happen for all usages, not just `.fromJson`

It could be a different mode where instead of:

```dart
MyClass(int foo): assert(foo != null);
```

we get:

```dart
MyClass(int foo) {
  if (foo == null) throw StateError('foo cannot be null');
}
```

mseijas: Yes! I considered that too. It just felt annoying that then we couldn't have `const` constructors anymore 😞 But I hear your point! In the end, and to keep things simpler, I thought that throwing errors in deserialization was more important because those models get initialized from data that comes from the server and we don't control. So the chances of something going wrong there seemed higher than the chances of manually building a model passing in invalid values.

mseijas:
> Maybe what you want is instead to change the default behavior of @JsonKey to match @required and @nullable?

Yes, exactly! That's what I want to do. 

Basically, I just want to depend on using the `@freezed` annotations and not have to duplicate those definitions with `@JsonKey` annotations as well.

So, if I use a `@required` annotation, I don't want to have to manually also write `@JsonKey(required: true, disallowNullValue: true)`

The same thing with the `@nullable` annotation, I don't want to have to manually also add `@JsonKey(required: false, disallowNullValue: false)`

Additionally, I thought it would also be nice that I didn't have `@freezed` and `@JsonKey` annotations that contradict themselves, like:
`@Default(42) @JsonKey(defaultValue: 30)`

Does that make sense? I'm happy to explore other ways of achieving this, but that was what I was aiming for.

packages/freezed/test/json_test.dart
rrousselGit: If the exception happens during deserialisation, const constructors don't matter

The benefit is, this continues to work on copyWith too

packages/freezed/test/json_test.dart
mseijas: But don't you think that this is problematic when working in a big team, and looking at someone else's code that you're not familiar with?

Although the behavior might be fixed based on what happens at run time, when reading this code it is not immediately clear which one of those decorators should superseed the other. I think it might be nice to prevent users from defining decorators that could in theory contradict each other, which means that one of them would actually end up being ignored.

rrousselGit: To me, it doesn't matter, as people shouldn't need to specific `@JsonKey` to begin with.

The support for `@JsonKey` in parameters is here only for people want to override the default behavior. So in that situation, it is clear that it takes over everything else.

What you want is not to check for inconsistencies. You want to change the default behavior.

packages/freezed/test/json_test.dart
mseijas: Ok, I'm happy to explore that approach instead if you think that's better!

packages/freezed/test/json_test.dart
mseijas: But there's also the case where you might want to specify a `@JsonKey` just so that you get a `CheckedFromJsonException` instead of a runtime assertion error. (Although if we modify our constructors the way you suggested above then this wouldn't matter anymore)

I do think it might still be nice to throw a `InvalidGenerationSourceError` during codegen when you have defined decorators that contradict each other as a sanity check? What do you think of that?

rrousselGit:
We could change the default `@JsonKey` such that if the property is nullable,  it is generated as `@JsonKey(nullable: true)`

I'm not quite sure why we would want `required` or `disallowNullValue` instead.

packages/freezed/test/json_test.dart
rrousselGit: > I do think it might still be nice to throw a InvalidGenerationSourceError during codegen when you have defined decorators that contradict each other as a sanity check? What do you think of that?

I would agree if people were expected to write `@JsonKey(something: true)` themselves. But they are not
As I mentioned, `@JsonKey` on parameters is only here for edge-cases to let people have fine grained control over the annotation.

mseijas: > As I mentioned, @JsonKey on parameters is only here for edge-cases to let people have fine grained control over the annotation.

Of course, yeah. I just meant to prevent errors from slipping through when dealing with these edge-cases.

mseijas:
> We could change the default `@JsonKey` such that if the property is nullable, it is generated as `@JsonKey(nullable: true)`
> 
> I'm not quite sure why we would want `required` or `disallowNullValue` instead.

So based on the json_serialization I think the `@JsonKey` properties should be set like this:

With freezed `@required` annotation, it should have the following `@JsonKey` settings:
`required: true` -> When true, generated code for fromJson will verify that the source JSON map contains the associated key.
`disallowNullValue: true` -> If true, generated code will throw a DisallowedNullValueException if the corresponding key exists, but the value is null.

With freezed `@nullable` annotation, it should have the following `@JsonKey` settings:
`required: false` -> When false, won't verify that source JSON map contains the associated key.

`nullable: true` works strangely actually, check out the documentation for it:
When true, generated code for fromJson will verify that the source JSON map contains the associated key.

If the key does not exist, a MissingRequiredKeysException exception is thrown.

mseijas:
That said, would you prefer for me to explore the approach of updating our constructors to throw any necessary exceptions instead of this other technique of creating/updating the `@JsonKey` settings?

rrousselGit:
> That said, would you prefer for me to explore the approach of updating our constructors to throw any necessary exceptions instead of this other technique of creating/updating the `@JsonKey` settings?

That would be preferable yes.

Ideally, the default behavior should stay as-is. And then, we'd have a setting on the `build.yaml` file & on `@Freezed()` to change the behavior.

mseijas:
> > That said, would you prefer for me to explore the approach of updating our constructors to throw any necessary exceptions instead of this other technique of creating/updating the `@JsonKey` settings?
> 
> That would be preferable yes.
> 
> Ideally, the default behavior should stay as-is. And then, we'd have a setting on the `build.yaml` file & on `@Freezed()` to change the behavior.

Cool! I'll explore that option 👍

mseijas:
Something to consider for later, I think we might still want to at least print a warning when doing something like this:

```dart
factory User({@Default(42) @JsonKey(defaultValue: 30) int age})
```

Since doing that really doesn't make sense. Right?

rrousselGit:
Warnings would be fine. But it's not the role of the code-generator.

We need a custom lint rule for this instead. So that they can be ignored or
disabled.

This is feasible, but is a different project

mseijas:
> We need a custom lint rule for this instead. So that they can be ignored or
disabled.

Oh I like that!

rrousselGit:
After I publish a project I'm working on, I'll try and make a custom linter
for all my different projects (flutter_hooks, freezed, provider, ...)

So this may come in due time. But it is a different issue

Le jeu. 14 mai 2020 à 17:47, Matias Seijas <notifications@github.com> a
écrit :

rrousselGit:
Do you still plan on working on this, or should we close this PR?

mseijas:
Yes! I haven't gotten around to it yet, but I will soon. Happy to open a different PR if you prefer though.

Issue #118, opened 2020-03-23
Stack Overflow Error


SunlightBro:
As Models in the Backend can have many to many relationships, 
consider two classes:

```dart
@freezed
abstract class Person with _$Person {
  factory Person(
    Appointment appointment, 
    int userId,
  ) = _Person;

  factory Person.fromJson(Map<String, dynamic> json) => _$PersonFromJson(json);
}

@freezed
abstract class Appointment with _$Appointment {
  factory Appointment(
    DateTime date, 
    Person creator,
  ) = _Company;

  factory Appointment.fromJson(Map<String, dynamic> json) => _$AppointmentFromJson(json);
}
```

since 0.10.0 of freezed this will produce 
```
[FINE] freezed:freezed on lib/apis/appointmentAPI/model/appointment.dart:Running FreezedGenerator
[SEVERE] freezed:freezed on lib/apis/appointmentAPI/model/appointment.dart:
Error running FreezedGenerator
Stack Overflow
...
```

SunlightBro:
Commplete Error Log when running
`flutter packages run build_runner build --delete-conflicting-outputs --verbose`
on
```yaml
dependencies:
  freezed: ^0.10.3
```

```
dart:core                                                    _SimpleUri.pathSegments
package:source_gen/src/utils.dart 102:15                     packageToAssetUrl
package:source_gen/src/utils.dart 74:14                      normalizeUrl
package:source_gen/src/type_checker.dart 216:7               _MirrorTypeChecker._computed
package:source_gen/src/type_checker.dart 230:38              _MirrorTypeChecker.isExactly
package:source_gen/src/type_checker.dart 159:7               TypeChecker.isAssignableFrom
package:source_gen/src/type_checker.dart 164:7               TypeChecker.isAssignableFromType
package:source_gen/src/type_checker.dart 137:43              TypeChecker._annotationsWhere
dart:core                                                    Iterable.isEmpty
package:freezed/src/templates/concrete_template.dart 100:1   _extension#0.hasNullable
package:freezed/src/templates/concrete_template.dart 337:12  IsNullable.isNullable
package:freezed/src/templates/concrete_template.dart 376:25  new Property.fromParameter
package:freezed/src/freezed_generator.dart 374:24            FreezedGenerator._parseConstructorsNeedsGeneration
package:freezed/src/freezed_generator.dart 258:9             FreezedGenerator.parseElement
[...]

PR #177, opened 2020-05-19
prevent endless parsing on bidirectional models

SunlightBro:
fixes issue #118 Stack Overflow Error

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -164,7 +164,7 @@ $runtimeType(
 @immutable
 class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
   final _computeElementDataCache = <ClassElement, Data>{};
-  final _parsedElementCheckList = <ClassElement>[];
+  final _parsedElementCheckSet = <ClassElement>{};
 
   Data _computeElementDataFor(ParameterElement parameter) {
     final parameterTypeElement = parameter?.type?.element;
@@ -175,8 +175,8 @@ class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
 
     return _computeElementDataCache.putIfAbsent(classElement, () {
       if (!typeChecker.hasAnnotationOf(classElement)) return null;
-      if (_parsedElementCheckList.contains(classElement)) return null;
-      _parsedElementCheckList.add(classElement);
+      if (_parsedElementCheckSet.contains(classElement)) return null;
+      _parsedElementCheckSet.add(classElement);
       return parseElement(
         _GlobalData(hasDiagnostics: false, hasJson: false),
         classElement,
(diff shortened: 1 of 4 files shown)


packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -174,6 +175,8 @@ class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
 
     return _computeElementDataCache.putIfAbsent(classElement, () {
       if (!typeChecker.hasAnnotationOf(classElement)) return null;
+      if (_parsedElementCheckList.contains(classElement)) return null;
```
rrousselGit, line 178: That "return null" sounds fishy.

Maybe we should replace the putIfAbsent with a cache[key] = whatever; instead. And do nothing in this situation (whereas the other line actually assign null)

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -164,6 +164,7 @@ $runtimeType(
 @immutable
 class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
   final _computeElementDataCache = <ClassElement, Data>{};
+  final _parsedElementCheckList = <ClassElement>[];
```
rrousselGit, line 167: Should be a Set

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -182,7 +185,10 @@ class FreezedGenerator extends ParserGenerator<_GlobalData, Data, Freezed> {
   }
 
   @override
-  Data parseElement(_GlobalData globalData, Element rawElement) {
+  Data parseElement(_GlobalData globalData, Element rawElement, {bool init}) {
```
rrousselGit, line 188: I don't quite understand what that init flag does. Could you explain?

SunlightBro: `parseElement()` gets called recursively through `_parseConstructorsNeedsGeneration()` and `_computeElementDataFor()`. 
But I only want to clear this `_parsedElementCheckList`, when its called by `GeneratorForAnnotation.generate()`,

rrousselGit: Why clear the list to begin with?

packages/freezed/lib/src/freezed_generator.dart
SunlightBro: so something like this ?
```dart    
if (!typeChecker.hasAnnotationOf(classElement) || _parsedElementCheckSet.contains(classElement)) {
  return null;
} else {
  _parsedElementCheckSet.add(classElement);
  final data = parseElement(
    _GlobalData(hasDiagnostics: false, hasJson: false),
    classElement,
  );
  _computeElementDataCache[classElement] = data;
  return data;
}
```

rrousselGit:
Could you add a test where it's not a List<Appointment> and try to use deep copy on it?

We should be able to write:

person.copyWith.appointment.person.appointment.person(name: 'foo')

SunlightBro:
Ok I added a test.

rrousselGit:
Thanks!

PR #188, opened 2020-06-01
Add devtools utilities #187

### LINKED ISSUE #187: Generate a debug-only wait to visit all the properties of an object
For debugging purpose, if would be interesting to generate utilities for devtools.

One of such utilities would be the ability to convert the object into a `Map<String, Object>` (to not confuse with toJson) such that we can write:

```dart
@freezed
abstract class Company with _Company {
  factory Company({String name, Director director}) = _Company;
}

@freezed
abstract class Director with _Director {
  factory Director({String name}) = _Director;
}

void main() {
  final company = Company(name: 'Example', director: Director(name: 'John'));

  print(company.debugToMap); // { name: 'Example', director: Director(name: 'John')) }

  print(company.director.debugToMap); // { name: 'John' } 
}
```

Notice how this works even although `Company` and `Director` are not serializable.

rrousselGit:


rrousselGit:
Canceling this after discussion with google

PR #196, opened 2020-06-12
naive regex for generics referencing codegen types

maks:
This expands a little the extra processing for types that the analyzer has incorrectly classified as `dynamic` because they are defined in code generated files to include single level generics, primarily intended to handle the use case of `Maps` and `Lists` of freezed generated union types.

### DIFF
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -471,7 +471,7 @@ extension DefaultValue on ParameterElement {
 String parseTypeSource(VariableElement element) {
   var type = element.type?.getDisplayString();
 
-  if (type == null || type.contains('dynamic')) {
+  if (type == null || type == 'dynamic') {
     if (element.type?.element != null &&
         element.type.isDynamic &&
         element.type.element.isSynthetic) {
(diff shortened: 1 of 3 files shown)


maks:
Tests still work in progress, as I'm struggling a bit to come up with a test that will only pass with this code change and fail without it.

rrousselGit:
Thanks!
The code in the associated issue should be more than enough for the test

Add the class in the `integration` folder, run the generator, then make a test that uses the class and verifies that it works as expected

rrousselGit:
Thanks
Could you trim the test of the unneeded constructors?

Also add a test for Map and a custom MyClass<A, B, C>

maks:
Ah, well spotted @rrousselGit ! the regex I used is too simple to cope with multiple generic Types that map and the custom class could have.
I'll fix that up.

rrousselGit:
Another case would be generic typedefs

It should be fine but doesn't hurt testing them too.

maks:
@rrousselGit thank you for fixing up the support and adding the extra tests to get the PR merged, **very** much appreciated! 👍 
One small issue, I think you accidentally reverted my change for checking for generics of dynamics:
https://github.com/rrousselGit/freezed/pull/196/commits/50c66344c5c83beae1b0fc6c5fdd3191165e660b#diff-d56f908c70e62a7ec25d0128ce3c7485R474

which actually means that the fix is not being applied in the generated code. It's just a 1 liner to switch the `==` for a `contains`, should I do a new PR for it?

rrousselGit:
Ah it wasn't tested, so I thought it wasn't needed. Sure go ahead

PR #197, opened 2020-06-12
[feature] add support for union types to implement interfaces individually

long1eu:


README.md
```diff
@@ -494,6 +495,48 @@ abstract class Person with _$Person {
...
+one of those types to implement a interface. You can do that using the
+`@Implements` decorator. In this case `City` is implementing  
+`GeographicArea`.
+
+```dart
+@freezed
+abstract class Example with _$Example {
+  const factory Example.person(String name, int age) = Person;
+
+  @Implements(GeographicArea)
```
rrousselGit, line 510: Maybe show that we can use Implements multiple times

rrousselGit:
Should à @/With be added too?

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -94,6 +94,31 @@ ${copyWith.abstractCopyWithGetter}
 ''';
   }
 
+  String get _implementations {
+    final implementationTypes = constructor.decorators
+        .where((element) => element.startsWith('@Implements'))
```
rrousselGit, line 99: The parsing should be done outside templates, on Freezed_generator.dart
Potentially with error handling too.

packages/freezed_annotation/lib/freezed_annotation.dart
```diff
@@ -66,3 +66,47 @@ class Default {
...
+/// just fields you can fulfil the interface contract by adding them in the
+/// constructor of the union type. Keep in mind that if the interface defines a
+/// method or a getter, that you implement in the class, you need to use the
+/// [Custom getters and methods](#custom-getters-and-methods) instructions.
+class Implements {
+  const Implements(this.type) : stringType = null;
+
+  const Implements.fromString(this.stringType) : type = null;
+
+  final Type type;
```
rrousselGit, line 110: Instead, have the default constructor stringify the type.

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -94,6 +94,31 @@ ${copyWith.abstractCopyWithGetter}
...
+        .map((element) {
+          // the fromString constructor contains commas
+          final offset = element.contains('fromString') ? 2 : 1;
```
rrousselGit, line 102: Instead of parsing the source code, we can read the content of the decorator and read stringType.

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -94,6 +94,31 @@ ${copyWith.abstractCopyWithGetter}
...
+          return element.substring(
+              element.indexOf('(') + offset, element.length - offset);
```
rrousselGit, line 104: This doesn't handle the following case:

```dart
class Example<T> {
  @Implements.fromString('Foo<T>')
  factory... 
}
```

rrousselGit: Actually that may work.
But we'll want a test for it 😊

packages/freezed/test/integration/implements_decorator.dart
```diff
@@ -0,0 +1,38 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+import 'package:meta/meta.dart';
```
rrousselGit, line 2: We need a test with multiple implements on one constructor

rrousselGit:
Thanks a lot for your work!

rrousselGit:
Thanks again! :)

Issue #89, opened 2020-03-05
Lists for union types have missing type in generated code


maks:
Given a model class which has a list of a "union subtype":
```dart
@freezed
abstract class FApp with _$FApp {
  factory FApp({Header head, List<Page> pages}) = _FApp;
}
```
where the union is:
```dart
@freezed
abstract class WidgetType with _$WidgetType {
  const factory WidgetType.page({Body body, Header header}) = Page;
  const factory WidgetType.body() = Body;
  const factory WidgetType.header({String title}) = Header;
}
```
the type for the list in the generated code is `dynamic` instead of the actual type:
```dart
mixin _$FApp {
  Header get head;
  List<dynamic> get pages;

  FApp copyWith({Header head, List<dynamic> pages});
}
```
but in that generated code I expected page getter to be `List<Page>` not `List<dynamic>`.

It all works fine if I use `List<WidgetType>` but I really want to ensure I'm getting a List of `Pages` and not just any `WidgetType`

I've got [a repo with the full demo code](https://github.com/maks/esky)

rrousselGit:
Do you still have the issue if you extract `FApp` in a different file?

maks:
Sorry I didn't explain it well, in my example repo FApp is in a separate file.

rrousselGit:
This was fixed as part of 10.0.3 but I forgot to close the issue

maks:
@rrousselGit I'm sorry if I'm doing something wrong, but with [my demo repo](https://github.com/maks/esky) that I tested with using freezed 0.10.9 now and I am still seeing the same problem, the generated code still has:
```dart
mixin _$FApp {
  Header get head;
  List<dynamic> get pages;

  $FAppCopyWith<FApp> get copyWith;
}
``` 
so still a `dynamic` List instead of the actual type from the freezed "union" that I defined 😞

rrousselGit:
What is Page then?

maks:
Sorry, Page is from the "union" I have:

```dart
import 'package:freezed_annotation/freezed_annotation.dart';

part 'widget_type.freezed.dart';

@freezed
abstract class WidgetType with _$WidgetType {
  const factory WidgetType.page({Body body, Header header}) = Page;
  const factory WidgetType.body() = Body;
  const factory WidgetType.header({String title}) = Header;
}
```

rrousselGit:
Oh I didn't realize, my bad. 

Indeed this is not supported.

maks:
[...]

PR #204, opened 2020-06-16
fix detecting generated generic for dynamics

maks:
Should I also add a (integration?) test for this? as currently nothing picks up that this didn't actually change the generated code as expected.

Fixes: #89

### DIFF
--- a/packages/freezed/test/multiple_constructors_test.dart
+++ b/packages/freezed/test/multiple_constructors_test.dart
@@ -280,14 +280,14 @@ void main() {
     );
   });
   group('NestedList', () {
-    test('does not generate dynamic', () async {
+    test('generates List of correct type', () async {
       final nestedListClass = await _getClassElement('ShallowNestedList');
 
       expect(nestedListClass.getField('children').type.getDisplayString(),
           'List<LeafNestedListItem>');
     });
 
-    test('does not generate dynamic for deep nested case', () async {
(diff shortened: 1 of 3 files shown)


rrousselGit:
Yeah we'll need a test for this.

maks:
@rrousselGit no worries, I'll add a test shortly

Matsue:
@maks 
If it is not annoying you, please use those tests. Those tests are failed before this change and passed after this change. 

https://github.com/maks/freezed/pull/1

maks:
Big thank you to @Matsue for adding tests for the fix.

packages/freezed/test/multiple_constructors_test.dart
```diff
@@ -289,4 +279,57 @@ void main() {
...
+  group('NestedMap', () {
+    test('generates Map of correct type', () async {
+      final nestedMapClass = await _getClassElement('ShallowNestedMap');
+
+      expect(nestedMapClass.getField('children').type.getDisplayString(),
+          'Map<String, LeafNestedMapItem>');
+    });
+
+    test('generates Map of correct type for deeply nested case', () async {
+      final nestedMapClass = await _getClassElement('DeepNestedMap');
```
rrousselGit, line 311: _getClassElement is very expensive. It shouldn't be called more than once per test.

rrousselGit:
Thanks!

Issue #162, opened 2020-04-29
Generated Methods not generated in class.freezed.dart file


noorforit:
Hi, I have tried to use freezed in my project. I followed the freezed documentation on pub.dev. Everything went fine but when I tried to access the generated methods on my Union class I found out that there are no such methods in .freezed.dart file.
```
@freezed
abstract class LoginEvent with _$LoginEvent {
  const factory LoginEvent.logIn(String email, String password) = LogIn;
}
```
I have declared the class using above mentioned syntax and then ran `flutter pub run build_runner watch` command and It generated .freezed.dart file with following contents:
```
// GENERATED CODE - DO NOT MODIFY BY HAND
// ignore_for_file: deprecated_member_use_from_same_package, use_function_type_syntax_for_parameters, unnecessary_const, avoid_init_to_null, invalid_override_different_default_values_named

part of 'login_bloc.dart';

// **************************************************************************
// FreezedGenerator
// **************************************************************************

T _$identity<T>(T value) => value;

class _$LoginEventTearOff {
  const _$LoginEventTearOff();

  LogIn logIn(String email, String password) {
    return LogIn(
      email,
      password,
    );
  }
}

// ignore: unused_element
const $LoginEvent = _$LoginEventTearOff();

mixin _$LoginEvent {
  String get email;
  String get password;

  $LoginEventCopyWith<LoginEvent> get copyWith;
}

abstract class $LoginEventCopyWith<$Res> {
  factory $LoginEventCopyWith(
          LoginEvent value, $Res Function(LoginEvent) then) =
      _$LoginEventCopyWithImpl<$Res>;
  $Res call({String email, String password});
}

class _$LoginEventCopyWithImpl<$Res> implements $LoginEventCopyWith<$Res> {
  _$LoginEventCopyWithImpl(this._value, this._then);

  final LoginEvent _value;
  // ignore: unused_field
  final $Res Function(LoginEvent) _then;

  @override
  $Res call({
    Object email = freezed,
    Object password = freezed,
  }) {
    return _then(_value.copyWith(
      email: email == freezed ? _value.email : email as String,
      password: password == freezed ? _value.password : password as String,
    ));
  }
}

abstract class $LogInCopyWith<$Res> implements $LoginEventCopyWith<$Res> {
  factory $LogInCopyWith(LogIn value, $Res Function(LogIn) then) =
      _$LogInCopyWithImpl<$Res>;
  @override
  $Res call({String email, String password});
}

[...]

PR #206, opened 2020-06-17
Support single-case unions

andrzejchm:
This pull-request allows for creating single-case unions. i.e:
```dart
@freezed
abstract class SingleCaseUnionWithData with _$SingleCaseUnionWithData {
  const factory SingleCaseUnionWithData.singleCase(int a, String b) = SingleCaseWithData;
}

@freezed
abstract class SingleCaseUnionWithoutData with _$SingleCaseUnionWithoutData {
  const factory SingleCaseUnionWithoutData.singleCase() = SingleCaseWithoutData;
}
```
so that you can call `map`, `when`, `maybeMap` and `maybeWhen`:

```dart

    singleCaseUnionWithDataExample.map(
      singleCase: (SingleCaseWithData singleCase) => '$singleCase',
    );

    singleCaseUnionWithDataExample.maybeWhen(
      singleCase: (a, b) => '$a $b', // this is optional
      orElse: () => 'this will not be called, since all cases are covered',
    );

    singleCaseUnionWithDataExample.maybeMap(
      singleCase: (SingleCaseWithData a) => '$a', // this is optional
      orElse: () => 'this will not be called, since all cases are covered',
    );

    singleCaseUnionWithoutDataExample.map(
      singleCase: (SingleCaseWithoutData singleCase) => '$singleCase',
    );

    singleCaseUnionWithoutDataExample.maybeWhen(
      singleCase: () => 'has no data', // this is optional
      orElse: () => 'this will not be called, since all cases are covered',
    );

    singleCaseUnionWithoutDataExample.maybeMap(
      singleCase: (SingleCaseWithoutData a) => '$a', // this is optional
      orElse: () => 'this will not be called, since all cases are covered',
    );

```

Addresses #162

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+packages/freezed/.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-packages/freezed/.idea
+.idea/
--- a/packages/freezed/example/lib/main.dart
+++ b/packages/freezed/example/lib/main.dart
@@ -18,16 +18,6 @@ abstract class Union with _$Union {
 }
 
 @freezed
-abstract class SingleCaseUnionWithData with _$SingleCaseUnionWithData {
-  const factory SingleCaseUnionWithData.singleCase(int a, String b) = SingleCaseWithData;
-}
-
-@freezed
-abstract class SingleCaseUnionWithoutData with _$SingleCaseUnionWithoutData {
-  const factory SingleCaseUnionWithoutData.singleCase() = SingleCaseWithoutData;
-}
-
-@freezed
 abstract class SharedProperty with _$SharedProperty {
   factory SharedProperty.person({String name, int age}) = SharedProperty0;
   factory SharedProperty.city({String name, int population}) = SharedProperty1;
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -0,0 +1 @@
+packages/freezed/.idea
```
rrousselGit, line 1: nit: .idea/

packages/freezed/example/lib/main.dart
```diff
@@ -89,8 +99,72 @@ void main() {
   // nice toString
   print(const Union(42)); // Union(value: 42)
   print(const Union.loading()); // Union.loading()
-  print(const Union.error(
-      'Failed to fetch')); // Union.error(message: Failed to fetch)
+  print(const Union.error('Failed to fetch')); // Union.error(message: Failed to fetch)
+
+  // union with single case - with data
+  const singleCaseUnionWithDataExample = SingleCaseUnionWithData.singleCase(43, 'string');
```
rrousselGit, line 105: Do we really need to add this to `freezed/example/lib/main.dart`? It sounds redundant with the previous part.

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -214,7 +214,7 @@ void debugFillProperties(DiagnosticPropertiesBuilder properties) {
   }
 
   String get _maybeMap {
-    if (allConstructors.length < 2) return '';
+    if (allConstructors.where((element) => element.name != null && element.name.isNotEmpty).isEmpty) return '';
```
rrousselGit, line 217: Empty names are supporter for unions. This leads to:

```dart
example.when(
  (value) {},
  named: (...) {},
);
```

packages/freezed/test/single_class_constructor_test.dart
```diff
@@ -172,6 +172,47 @@ void main() {
     expect('${SingleNamedCtor.named(42)}', 'SingleNamedCtor.named(a: 42)');
   });
 
+  test('single-case union does have map', () async {
+    await expectLater(compile(r'''
```
rrousselGit, line 176: Instead of compile/completes, use the `map` method normally.
The current test assert that there is a `map` member, but not what it is and what it does.

andrzejchm:
thanks for the feedback @rrousselGit ! I've applied fixes

rrousselGit:
I've added some tests on `master` that I forgot to make before.
This should break this PR

As I mentioned, we cannot have a `if name.isNotEmpty` to generate `when`.

andrzejchm:
@rrousselGit not sure where the problem is actually. with the proposed changes, `when` `map` etc. will be generated for all classes that have at least one named constructor. since a  class with just single no-name constructor is not considered a union, right? so a class like

```dart
@freezed
abstract class Union with _$Union {
  const factory Union(int value) = Data;
  const factory Union.loading() = Loading;
}
```
will correctly produce `when`, `map` etc methods:
```dart
union.when(
  (value) {},
  loading: (...) {},
);
```

The actual condition that I wrote can be broken down to:
```dart
final namedConstructors = allConstructors.where((element) => element.name != null && element.name.isNotEmpty);
if (namedConstructors.isEmpty) return ''; // don't generate `map`, `when`. etc if class does not have any named constructor.
else return '...' // generate `map`, `when` etc.
```

rrousselGit:
Oh... Indeed my bad. I didn't think about the fact that we can only have one unnamed constructor, so it's fine.

Issue #220, opened 2020-06-29
DiagnosticableTreeMixin related code still generated when foundation.dart is not imported


skybur:
I have this freezed class which uses `TimeOfDay` from `material` library.

```dart
import 'package:flutter/material.dart';
import 'package:freezed_annotation/freezed_annotation.dart';

part 'time_slot.freezed.dart';

@freezed
abstract class TimeSlot with _$TimeSlot {
  factory TimeSlot({
    TimeOfDay start,
    TimeOfDay end,
  }) = _TimeSlot;
}
```

The generated class still contains `DiagnosticableTreeMixin` even though I didn't add import for `flutter/foundation.dart`, thus making a compile error.

Freezed version is `0.10.9`

rrousselGit:
Probably an issue with the logic that determines if "foundations" is imported.

Fancy making a PR? Shouldn't be too difficult, but I have other things to do.

skybur:
I have no experience dealing with code generation but I'll see what I can do.

DFelten:
Same issue here. I had to manual import foundation within the freezed class.

PR #231, opened 2020-07-14
DiagnosticableTreeMixin related code still generated when foundation.dart is not imported

avbk:
Hey Remi,

first of all thanks for this awesome library :tada: 

I wanted to get a deeper understanding of the internals of freezed and thus I considered to fix a bug or two... So I picked #220, which you labeled as a good starting point. _(honestly, it was trickier than I expected it to be in the first place)_

The problem with the current implementation is that `_libraryHasElement` searches recursively for any exported classes within the imports, but it does not take the `show` and `hide` keyword into account. 
I moved `_libraryHasElement` into a separate class `RecursiveImportLocator` and added some logic to consider `show` and `hide`. It's quite a simple approach and I might have overlooked some edge case, but at least for the described issue it works :wink: 

I hope this PR is fine for you and I'm curious about your feedback :slightly_smiling_face: 

Best regards,
Alex

@skybur @rrousselGit

### DIFF
--- a/packages/freezed/example/lib/time_slot.dart
+++ b/packages/freezed/example/lib/time_slot.dart
@@ -1,0 +1,15 @@
+import 'package:flutter/material.dart';
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'time_slot.freezed.dart';
+
+@freezed
+
+/// This is class has been added to address the issue described in
+/// https://github.com/rrousselGit/freezed/issues/220
+abstract class TimeSlot with _$TimeSlot {
+  factory TimeSlot({
+    TimeOfDay start,
+    TimeOfDay end,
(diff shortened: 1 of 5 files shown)


avbk:
Sadly the decorator tests fail on travis eventhough they work on my machine... 
And I actually don't expect any regression in this part of freezed. Do you have a clue about what might be wrong?

rrousselGit:
It was a bug due to the CI using a different version of a dependency
If you pull from master, that should fix it

avbk:
Awesome, will take care of this tomorrow

avbk:
... and done :slightly_smiling_face:

rrousselGit:
Nice thanks a lot!

avbk:
Sure thing :slightly_smiling_face:

PR #233, opened 2020-07-16
Add links to the required dependencies

bassel:
This makes it easier to navigate to the packages and check their latest version when adding them to pubspec


rrousselGit:
Nice thanks!

PR #239, opened 2020-08-02
Improve pub points

Mravuri96:
Should comply with "Only use double quotes for strings containing single quotes." requirement, which was leading to a 10 point deduction to the pub score.


rrousselGit:
Thanks!

Issue #240, opened 2020-08-03
Not clear how to apply custom json serializable converter


SandroMaglione:
I am trying to apply my custom json converter. The README states that it is possible to add a custom json serializable converter to a model created with freezed, but it does not seem to provide an example on how to actually apply it to a model.

I tried different strategies, but all seem unsuccessful.
``` dart
// 1. Outside the model

@freezed
@ModelConverter()
abstract class Model with _$Model {
  const factory Model({
    String name,
    String surname,
  }) = _Model;

  factory Model.fromJson(Map<String, dynamic> json) => _$ModelFromJson(json);
}
```
``` dart
// 2. On the factory constructor

@freezed
abstract class Model with _$Model {
  @ModelConverter()
  const factory Model({
    String name,
    String surname,
  }) = _Model;

  factory Model.fromJson(Map<String, dynamic> json) => _$ModelFromJson(json);
}
```
``` dart
// 3. On the factory fromJson method

@freezed
abstract class Model with _$Model {
  const factory Model({
    String name,
    String surname,
  }) = _Model;

  @ModelConverter()
  factory Model.fromJson(Map<String, dynamic> json) => _$ModelFromJson(json);
}
```
As stated, none of this solutions seem to work.

How can I apply my custom converter such that `toJson` and `fromJson` generated with json_serializable utilize my custom converter?

rrousselGit:
The decorator isn't placed on the `Model` class, but the places where `Model` is used:

```dart
@freezed
abstract class Another with _$Another {
  const factory Another({
    @ ModelConverter Model model,
  }) = _Another;

  factory Anotherl.fromJson(Map<String, dynamic> json) => _$AnotherFromJson(json);
}
```

SandroMaglione:
I see, now it works. Thanks.

I was having doubts because the `Model` inside `Another` class was contained in a `List`, so I thought it would not work.
``` dart
@freezed
abstract class Another with _$Another {
  const factory Another({
    // Adding the decorator with a list works as well!
    @ModelConverter List<Model> modelList,
  }) = _Another;

  factory Another.fromJson(Map<String, dynamic> json) => _$AnotherFromJson(json);
}
```

Do you think it would be useful to add this example to the docs (both for the plain `Model` and `List` examples)? Should I open a PR?

rrousselGit:
Sure, feel free to make a PR

Issue #232, opened 2020-07-15
toJson/fromJson is broken for nested lists of freezed objects


vishna:
Consider the following example:

```dart
import 'package:freezed_annotation/freezed_annotation.dart';

part 'main.freezed.dart';
part 'main.g.dart';

@freezed
abstract class TodoList with _$TodoList {
  factory TodoList(List<Todo> todos) = _TodoList;
  factory TodoList.fromJson(Map<String, dynamic> json) => _$TodoListFromJson(json);
}

@freezed
abstract class Todo with _$Todo {
  factory Todo(String name, bool checked) = _Todo;
  factory Todo.fromJson(Map<String, dynamic> json) => _$TodoFromJson(json);
}
```

Observe the following test fail:
```dart
import 'package:flutter_test/flutter_test.dart';

import 'package:freezed_json_bug/freezed_json_bug.dart';
import 'package:collection/collection.dart';

void main() {
  test('serialize deserialize', () {
    final todos = <Todo>[
      Todo("Item 1", true),
      Todo("Item 2", false),
    ];

    final todoList = TodoList(todos);

    final todoListJson = todoList.toJson();

    final todoListReserialized = TodoList.fromJson(todoListJson);

    Function deepEq = const DeepCollectionEquality().equals;

    expect(deepEq(todoList.toJson(), todoListReserialized.toJson()), true);
  });
}
```

with the following stacktrace:

```
type '_$_Todo' is not a subtype of type 'Map<String, dynamic>' in type cast
package:freezed_json_bug/src/main.g.dart 13:55        _$_$_TodoListFromJson.<fn>
dart:_internal                                        ListIterable.toList
package:freezed_json_bug/src/main.g.dart 14:11        _$_$_TodoListFromJson
package:freezed_json_bug/src/main.freezed.dart 89:7   new _$_TodoList.fromJson
package:freezed_json_bug/src/main.freezed.dart 12:20  _$TodoListFromJson
package:freezed_json_bug/src/main.dart 9:59           new TodoList.fromJson
test/freezed_json_bug_test.dart 17:43                 main.<fn>
```

This is because in `main.g.dart` we have:

```dart
Map<String, dynamic> _$_$_TodoListToJson(_$_TodoList instance) =>
    <String, dynamic>{
      'todos': instance.todos,
    };
```

rather than:

```dart
Map<String, dynamic> _$_$_TodoListToJson(_$_TodoList instance) =>
    <String, dynamic>{
[...]

PR #241, opened 2020-08-03
Added example of how to apply json custom converter in freezed documentation

SandroMaglione:
The PR aims to clarify two features from #240 and #232.

It adds two code examples on how to apply the custom converter decorator `MyResponseConverter` already present inside the documentation. Specifically, one example illustrates how to apply the decorator by passing it as a constructor parameter, while the second one shows that the converter works also on a `List` parameter.

It also adds a note below it to clarify that, in order to serialize nested list of freezer objects, you are supposed to either specify a `@JsonSerializable(explicitToJson: true)` or change `explicit_to_json` inside your `build.yaml` file (taken from [here](https://github.com/rrousselGit/freezed/issues/232#issuecomment-658708471)).

**Note**: The PR introduces a repetition. The reference to `build.yaml` to define some custom json_serializable flags is present both inside the PR and at the very bottom of the original documentation. I think the note may be helpful right below the json_serializable section to clarify the concepts explained above.

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -1083,6 +1083,8 @@ abstract class MyModel with _$MyModel {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


rrousselGit:
Thanks!

PR #250, opened 2020-08-16
Add Freezed extension in the Readme

Blaxou:
As discussed on Twitter, here is the pull request that adds the extension I built for VSCode in the ReadMe. About the section name in the README, I'm not convinced by my choice of "Utilities", so if you have a better idea lmk !


rrousselGit:
Nice, thanks!

PR #286, opened 2020-10-14
Support block body syntax for `fromJson` constructor.

zulfahmi93:
This pull request should fix the regression introduced by PR #283 which causes the `fromJson` constructor defined using the block body syntax no longer being generated. This PR should also enable the custom`fromJson` constructor implementation.

### DIFF
--- a/packages/freezed/CHANGELOG.md
+++ b/packages/freezed/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Fixed a bug where a `fromJson` constructor is not generated when using the
+  block body syntax.
+
 # 0.12.2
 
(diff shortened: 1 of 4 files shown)


zulfahmi93:
I think this should be documented properly in the README so that the library consumers know that this is not a supported scenario as I didn't find any reference stating about this. Maybe I missed it?

rrousselGit:
There is no mention of it being supported either

zulfahmi93:
We should explicitly state this in the documentation. As of our team, we're enforcing the use of block body syntax for all constructors and methods, although it is a simple constructor/method. Since previously the codegen works, we assumed that this is a supported scenario. Whenever we upgraded to the latest freezed package, all went haywire and we need to pin to previous version of freezed package.

Issue #303, opened 2020-11-17
Rename Result in generated code please


rmargolinross:
I have a local class called Result. This causes many issues due to Freezed using this keyword.

Solution:
It would be very nice if Freezed used a more generic keyword instead. Something maybe like TResult or to ensure no name collisions, maybe a FreezedResult or something.

PR #304, opened 2020-11-17
rename Result to TResult to avoid naming collision issues

### LINKED ISSUE #70: Support import alias
Hi,
simple code:

```dart
import 'package:dart_algorand/algod.dart' as algod;

@freezed
abstract class BaseState with _$BaseState {
  factory BaseState({algod.Account accountInfo}) = _BaseState;
}
```

It will not work as generated file will use just Account without prefix algod.

Hope it's clear enough.

a.

rmargolinross:
This will address #303 and possibly fix #70 in a sense.


rrousselGit:
Thanks!

PR #319, opened 2020-12-10
update package freezed_annotation to 0.12.0-nullsafety

unger1984:
Update first package. For update other, need publish freezed_annotation 0.12.0-nullsafety as pre-release


rrousselGit:
As much as I appreciate the effort, this is not something that we should merge/publish until Freezed itself is migrated

unger1984:
> As much as I appreciate the effort, this is not something that we should merge/publish until Freezed itself is migrated

Sure! It`s still not complete, but I think maybe this might be a good start

rrousselGit:
I'll close this as I fully migrated Freezed in a separate PR

Thanks for your effort!

PR #329, opened 2020-12-20
Update README for generic class mixin example.

pblinux:
In the "Mixin and Interfaces for individual classes for union types" section, the generic class mixin example is missing `_$'.


rrousselGit:
Oh thanks!

PR #333, opened 2020-12-28
add option to exclude code coverage from generated files 

lunaticcoding:
Add the option of excluding the generated files from the test code coverage.

Default behaviour remains the same

### DIFF
--- a/packages/freezed/lib/builder.dart
+++ b/packages/freezed/lib/builder.dart
@@ -5,9 +5,13 @@ import 'src/freezed_generator.dart';
 
 /// Builds generators for `build_runner` to run
 Builder freezed(BuilderOptions options) {
-  return PartBuilder([FreezedGenerator(options.config)], '.freezed.dart',
-      header: '''
-// GENERATED CODE - DO NOT MODIFY BY HAND
-// ignore_for_file: deprecated_member_use, deprecated_member_use_from_same_package, use_function_type_syntax_for_parameters, unnecessary_const, avoid_init_to_null, invalid_override_different_default_values_named, prefer_expression_function_bodies
-    ''');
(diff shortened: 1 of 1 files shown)


rrousselGit:
This isn't the responsibility of Freezed.

Instead, use the a custom `builder.yaml` to insert such line.

lunaticcoding:
@rrousselGit could you provide a minimal example? 

```
targets:
    $default:
        builders:
            freezed: 
                 ??
```

I can only find 
- dev_options: Map, Optional: A free-form map which will be passed to the Builder as a BuilderOptions when it is constructed. Usage varies depending on the particular builder. The values in this map override Builder defaults or non mode-specific options per-key when the build is done in dev mode.
- release_options: Map, Optional: A free-form map which will be passed to the Builder as a BuilderOptions when it is constructed. Usage varies depending on the particular builder. The values in this map override Builder defaults or non mode-specific options when the build is done in release mode.

to supply the flag :s

lunaticcoding:
@rrousselGit after reading through the docs, I can't find a way to add a line at the beginning of a generated file. I can only specify arguments that the corresponding plugin has to handle then. 

Could you please show me how I can do this? Because I don't see how this would be possible.

rrousselGit:
Sorry for the delay

Mad bad I confused `ignore_for_file` with the ability to add _anything_ on top of files

But I still believe this should be the role of `source_gen` instead, not Freezed, as this isn't Freezed-specific. You could make a PR to `source_gen` that introduce a `header` config similar to `ignore_for_file`, that all generators (including Freezed) would benefit from.

rrousselGit:
Closing as it should be the responsibility of a different package.

PR #354, opened 2021-01-27
Replace explicit_to_json with explicitToJson

robertpiosik:
`explicit_to_json` isn't defined on `JsonSerializable` class.


rrousselGit:
Thanks!

PR #364, opened 2021-02-18
Get rid of the constraint of freezed classes having to be abstract

gaetschwartz:
As discussed in Twitter DMs, here is the PR.

All tests pass except one that I removed because it was explicitly check if the `@freezed` class was abstract which is not a constraint anymore.

### DIFF
--- a/packages/freezed/CHANGELOG.md
+++ b/packages/freezed/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 0.13.0-nullsafety.1
+
+- Fixed a bug causing properties with annotations to generate the annotations _before_
+  the `required keyword, leading to a compilation error (see #351).
+
 # 0.13.0-nullsafety.0
 
 Migrated to null safety
--- a/packages/freezed/CHANGELOG.md
+++ b/packages/freezed/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.13.0-nullsafety.2
+
+- Excluded getters and late variables from the generated `toString` as they could potentially cause a StackOverflow
+
 # 0.13.0-nullsafety.1
 
 - Fixed a bug causing properties with annotations to generate the annotations _before_
--- a/packages/freezed/CHANGELOG.md
+++ b/packages/freezed/CHANGELOG.md
@@ -1,6 +1,39 @@
 # 0.14.0-nullsafety.0
 
-- Classes do not need to be declared as `abstract` anymore. Allows to use mixins and/or implement/extend other classes and have the IDE give meaningful feedback.
+- Freezed classes no-longer need to be abstract.  
+  Before:
+  ```dart
+  @freezed
+  abstract class Example with _$Example {
+    factory Example(int value) = _Example;
+  }
+  ```
+
+  after:
+
+  ```dart
+  @freezed
+  abstract class Example with _$Example {
+    factory Example(int value) = _Example;
+  }
+  ```
+
+  This leads to better error messages when a Freezed class uses interfaces/mixins
+  but the class is missing a specific property.
+
+- It is now allowed to add unimplemented getter to Freezed classes.
+  This can be useful to guarantee that union-types all have a common property:
+
+  ```dart
+  @freezed
+  class Union with _$Union {
+    const factory Union.left(int value) = _Left;
+    const factory Union.right(int value) = _Left;
+
+    @override
+    int get value; // adding this forces all union cases to possess a `value` property
+  }
+  ```
 
 # 0.13.0-nullsafety.2
 
--- a/packages/freezed/CHANGELOG.md
+++ b/packages/freezed/CHANGELOG.md
@@ -13,7 +13,7 @@
 
   ```dart
   @freezed
-  abstract class Example with _$Example {
+  class Example with _$Example {
     factory Example(int value) = _Example;
   }
   ```
--- a/packages/freezed/CHANGELOG.md
+++ b/packages/freezed/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 0.14.0-nullsafety.0
+
+- Classes do not need to be declared as `abstract` anymore. Allows to use mixins and/or implement/extend other classes and have the IDE give meaningful feedback.
+
 # 0.13.0-nullsafety.2
 
 - Excluded getters and late variables from the generated `toString` as they could potentially cause a StackOverflow
--- a/packages/freezed/lib/builder.dart
+++ b/packages/freezed/lib/builder.dart
@@ -10,6 +10,6 @@ Builder freezed(BuilderOptions options) {
   return PartBuilder([FreezedGenerator(options.config)], '.freezed.dart',
       header: '''
 // GENERATED CODE - DO NOT MODIFY BY HAND
-// ignore_for_file: unused_element, deprecated_member_use, deprecated_member_use_from_same_package, use_function_type_syntax_for_parameters, unnecessary_const, avoid_init_to_null, invalid_override_different_default_values_named, prefer_expression_function_bodies
+// ignore_for_file: unused_element, deprecated_member_use, deprecated_member_use_from_same_package, use_function_type_syntax_for_parameters, unnecessary_const, avoid_init_to_null, invalid_override_different_default_values_named, prefer_expression_function_bodies, annotate_overrides
     ''');
 }
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -102,6 +102,12 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
     @required bool shouldUseExtends,
   }) {
     // TODO: verify _$name is mixed-in
+    if (element.isAbstract) {
+      print(
+        'Declared @freezed ${element.name} class with abstract, but it is not need anymore.\n'
+        'Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#the-abstract-keyword',
+      );
+    }
 
     // Invalid constructors check
     for (final constructor in element.constructors) {
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -423,7 +423,6 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
         commonProperties: commonProperties,
         genericsDefinition: data.genericsDefinitionTemplate,
         genericsParameter: data.genericsParameterTemplate,
-        concretePropertiesName: data.concretePropertiesName,
         copyWith: CopyWith(
           clonedClassName: constructor.redirectedName,
           cloneableProperties: constructor.cloneableProperties,
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -103,9 +103,11 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
   }) {
     // TODO: verify _$name is mixed-in
     if (element.isAbstract) {
-      print(
-        'Declared @freezed ${element.name} class with abstract, but it is not need anymore.\n'
-        'Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#the-abstract-keyword',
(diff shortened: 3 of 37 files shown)


gaetschwartz:
It appears it also added an old commit to the PR which has already been merged if I'm not being mistaken. Anyway it doesn't change any current files.

packages/freezed/lib/builder.dart
```diff
@@ -10,6 +10,6 @@ Builder freezed(BuilderOptions options) {
   return PartBuilder([FreezedGenerator(options.config)], '.freezed.dart',
       header: '''
 // GENERATED CODE - DO NOT MODIFY BY HAND
-// ignore_for_file: unused_element, deprecated_member_use, deprecated_member_use_from_same_package, use_function_type_syntax_for_parameters, unnecessary_const, avoid_init_to_null, invalid_override_different_default_values_named, prefer_expression_function_bodies
+// ignore_for_file: unused_element, deprecated_member_use, deprecated_member_use_from_same_package, use_function_type_syntax_for_parameters, unnecessary_const, avoid_init_to_null, invalid_override_different_default_values_named, prefer_expression_function_bodies, annotate_overrides
```
rrousselGit, line 13: why is this lint disabled?

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -102,12 +102,6 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
     @required bool shouldUseExtends,
   }) {
     // TODO: verify _$name is mixed-in
-    if (!element.isAbstract) {
```
rrousselGit, line 105: Do we want to reverse the exception? Throw if the class is abstract

packages/freezed/lib/src/templates/abstract_template.dart
```diff
@@ -49,27 +51,26 @@ ${copyWith.commonContreteImpl(abstractProperties)}
...
     if (!shouldGenerateJson) return '';
-
-    return 'Map<String, dynamic> toJson();';
+    return 'Map<String, dynamic> toJson() => throw $privConstUsedErrorVarName;';
   }
 
   String get _when {
     if (!allConstructors.shouldGenerateUnions) return '';
-    return '${whenPrototype(allConstructors)};';
+    return '${whenPrototype(allConstructors)} => throw $privConstUsedErrorVarName;';
```
rrousselGit, line 59:
```suggestion
    return '${whenPrototype(allConstructors)} => throw UnimplementedError();';
```

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -362,7 +356,8 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
 
   @override
   Iterable<Object> generateForAll(GlobalData globalData) sync* {
-    yield r'T _$identity<T>(T value) => value;';
+    yield r'T _$identity<T>(T value) => value;'
+        '\n\nfinal $privConstUsedErrorVarName = UnsupportedError(\'$privConstUsedErrorString\');';
```
rrousselGit, line 360: I'm not fond of this variable, nor do I like "UnsupportedError"

rrousselGit: Actually, keep it this way

rrousselGit, line 360: Use a separate `yield` instead

packages/freezed/lib/builder.dart
gaetschwartz: Because when using mixins, the class overrides the mixxins abstract getters as expected but the linter is unhappy because it wants us to annotate it with `@override`. If I'm not mistaken `@override` is only to give aesthetic and doesn't actually change the behavior of the method right ?

packages/freezed/lib/src/freezed_generator.dart
gaetschwartz: I thought about it but that would be breaking. Keeping as abstract doesn't break anything at all so we could maybe just print a warning advising people to get rid of the `abstract` ?

rrousselGit: A print is fine. Make sure to test it too 🙂

packages/freezed/lib/builder.dart
rrousselGit: I'd still add the necessary `@override` 

I try to avoid disabling lints, so that tests can pick up warnings inside generated code to avoid potential issues.

packages/freezed/lib/src/templates/properties.dart
```diff
@@ -9,6 +9,16 @@ import '../utils.dart';
 import 'concrete_template.dart';
 import 'prototypes.dart';
 
+const privConstUsedErrorString =
+    'It seems like you constructed your class using MyClass._()'
+    ' or by manually adding another constructor. '
```
rrousselGit, line 14: It isn't possible to add other constructors, so this mention is not necessary

packages/freezed/lib/src/templates/properties.dart
```diff
@@ -9,6 +9,16 @@ import '../utils.dart';
 import 'concrete_template.dart';
 import 'prototypes.dart';
 
+const privConstUsedErrorString =
+    'It seems like you constructed your class using MyClass._()'
```
rrousselGit, line 13: Could we make `MyClass` the actual class name?

packages/freezed/lib/src/templates/abstract_template.dart
```diff
@@ -49,27 +51,26 @@ ${copyWith.commonContreteImpl(abstractProperties)}
...
   String get _map {
     if (!allConstructors.shouldGenerateUnions) return '';
-    return '${mapPrototype(allConstructors, genericsParameter)};';
+    return '${mapPrototype(allConstructors, genericsParameter)} => throw $privConstUsedErrorVarName;';
   }
 
   String get _maybeMap {
     if (!allConstructors.shouldGenerateUnions) return '';
-    return '${maybeMapPrototype(allConstructors, genericsParameter)};';
+    return '${maybeMapPrototype(allConstructors, genericsParameter)} => throw $privConstUsedErrorVarName;';
```
rrousselGit, line 74: we need tests for all of those exceptions.

packages/freezed/lib/src/templates/properties.dart
gaetschwartz: Alright, removing it tonight.

packages/freezed/lib/src/templates/properties.dart
gaetschwartz: Sure, just figured that it would be easier and more performant to have it like this but a method like this would also do.
```dart
String _errorMessage(String s) => "Text here $s text here";
```

packages/freezed/lib/src/templates/abstract_template.dart
gaetschwartz: True, doing it tonight also.

packages/freezed/lib/src/freezed_generator.dart
gaetschwartz: I honestly have no idea how to test it, will try to look into it.

packages/freezed/lib/src/templates/properties.dart
gaetschwartz: By looking at it it would over complicate a lot the code, having to pass the class name down the tree up to the `Getter ` class, plus I can't seem to be able to get a consistent name with it. We can just not mention the class if you want ?

packages/freezed/lib/builder.dart
gaetschwartz: That would need a lot more work for not much. Especially when `@override` is purely optional so I don't know. We'd have to check all `MyClass`'s super classes and compute which ones are being overridden.

packages/freezed/lib/src/freezed_generator.dart
gaetschwartz: I can't find a proper way to do so, any idea ?

rrousselGit: See source_gen_test.dart

You can pass expected logs in its annotation

gaetschwartz: Yeah but the thing is it doesn't throw, just prints a warning. `source_gen_dest` doesn't seem to have api to check the logs. Is it *that* necessary to test ? It doesn't really change from the past behavior and it's just a warning ?

gaetschwartz (commented):
LGTM, is there anything left to do ?

packages/freezed/CHANGELOG.md
```diff
@@ -1,3 +1,40 @@
...
+  abstract class Example with _$Example {
+    factory Example(int value) = _Example;
+  }
+  ```
+
+  after:
+
+  ```dart
+  @freezed
+  abstract class Example with _$Example {
```
mono0926, line 16: Shouldn't `abstract` be omitted?

mono0926, line 16:
```suggestion
  class Example with _$Example {
```

Looks like a mistake 👀

rrousselGit:
We still need benchmarks to see if this doesn't increase bundle size or decrease read performance

gaetschwartz:
I can try and do that with an existing app using a lot of `freezed` classes. For benchmarks, I don't really know how to proceed, how would you ?

gaetschwartz:
Alright, so for an app with `52` complete `freezed` classes (> 5 fields each), here is the result.

|Version | Size|
|---|---|
|`0.14.x`|`812 KB`|
|`0.13.x`|`812 KB`| 

Analysis done using `flutter build apk --analyze-size --target-platform android-arm64`.

As you can see, it doesn't seem to alter size at all.

rrousselGit:
> I can try and do that with an existing app using a lot of `freezed` classes. For benchmarks, I don't really know how to proceed, how would you ?

You can look at Riverpod's benchmark folder

gaetschwartz:
@rrousselGit Interesting. This makes it actually faster: 

image


```dart
import 'common.dart';
import 'models/new/bidirectional.dart' as n;
import 'models/old/bidirectional.dart' as o;

const int _kNumIterations = 100000;
const int _kNumWarmUp = 1000;

void main() {
  assert(
    false,
    "Don't run benchmarks in checked mode! Use 'flutter run --release'.",
  );

  final printer = BenchmarkResultPrinter();
  const scale = 1000.0 / _kNumIterations;
  final watch = Stopwatch();
  final dateTime = DateTime(2021);

  final oldPerson = o.Person(name: 'Hello', age: 99);
  final newPerson = n.Person(name: 'Hello', age: 99);

  // Warm up lap
  for (var i = 0; i < _kNumWarmUp; i += 1) {
    copyNew(newPerson, dateTime);
    copyOld(oldPerson, dateTime);
  }

  watch.reset();
  for (var i = 0; i < _kNumIterations; i += 1) {
    watch.start();
    copyOld(oldPerson, dateTime);
    watch.stop();
  }

  final oldCopyWithTime = watch.elapsedMicroseconds;

  printer.addResult(
    description: 'copy_with_old',
    value: oldCopyWithTime * scale,
    unit: 'ns per iteration',
    name: 'copy_with_old_iteration',
  );

  watch.reset();
  for (var i = 0; i < _kNumIterations; i += 1) {
    watch.start();
    copyNew(newPerson, dateTime);
    watch.stop();
  }

  final newCopyWithTime = watch.elapsedMicroseconds;

  printer
    ..addResult(
      description: 'copy_with_new',
      value: newCopyWithTime * scale,
      unit: 'ns per iteration',
      name: 'copy_with_new_iteration',
    )
    ..printToStdout();
}

void copyNew(n.Person newPerson, DateTime dateTime) {
  newPerson.copyWith(
      age: 29,
      name: 'Foo',
      appointment: n.Appointment(
        title: 'Rdv',
        date: dateTime,
        creator: n.Person(
          name: 'Bar',
          age: 63,
        ),
      ));
}

void copyOld(o.Person oldPerson, DateTime dateTime) {
  oldPerson.copyWith(
      age: 29,
      name: 'Foo',
      appointment: o.Appointment(
        title: 'Rdv',
        date: dateTime,
        creator: o.Person(
          name: 'Bar',
          age: 63,
        ),
      ));
}

```

rrousselGit:
I'll merge this, thanks!

PR #369, opened 2021-02-28
Fixing @required annotation in README

amitkot:
@required annotation in README file was missing the @ sign

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -350,7 +350,7 @@ update your code to:
 abstract class Person with _$Person {
   const factory Person(
     String name, {
-    required int age,
+    @required int age,
     Gender? gender,
   }) = _Person;
 }


rrousselGit:
Hello!

This `@` is not missing. That is the correct syntax with null-safety, so I'll close this

rrousselGit:
>  in other places in the README required is used as an annotation.

Indeed, because other places use the pre-null-safety syntax. 
The context of this snippet is specifically about what the syntax would look like with null-safety.

> I think the null-safety documentation should include other changes as well, e.g. the requirement for wither adding @Default() or making the type nullable in optional named arguments.

I'm not sure what this means. Do you have an example?

amitkot:
> > in other places in the README required is used as an annotation.
> 
> Indeed, because other places use the pre-null-safety syntax.
> The context of this snippet is specifically about what the syntax would look like with null-safety.

I read it again and you are correct, sorry!

> > I think the null-safety documentation should include other changes as well, e.g. the requirement for wither adding @default() or making the type nullable in optional named arguments.
> 
> I'm not sure what this means. Do you have an example?

Sure. Using this code:
```dart
@freezed
abstract class Session implements _$Session {
  const factory Session({
    @JsonKey(ignore: true) String id,
    required String title,
    required String description,
  }) = _Session;
  const Session._();
```
I got this error:

    The parameter `id` of `SessionStage.` is non-nullable but is neither required nor marked with @Default

After adding a default value, it builds. New code:
```dart
@freezed
abstract class Session implements _$Session {
  const factory Session({
    @JsonKey(ignore: true) @Default('') String id,
    required String title,
    required String description,
  }) = _Session;
  const Session._();
```

PR #370, opened 2021-03-01

### LINKED ISSUE #331: genericArgumentFactories for freezed
JsonSerializable now has a `genericArgumentFactories` option that adds an extra param for `fromJson` and `toJson` but this param isn't added to the generated code, for example:

```dart
@freezed
abstract class Example<T> with _$Example<T> {
  @JsonSerializable(genericArgumentFactories: true)
  factory Example({
    String status,
    T value,
  }) = _Example;

  factory Example.fromJson(Map<String, dynamic> json) =>
      _$ExampleFromJson(json);
}
```

But now, `_$_$_ExampleToJson` and `_$_$_ExampleFromJson` have and extra param `toJsonT` and `fromJsonT` that is not being sent.

Using the `fromJson` and `toJson` functions from JsonKey, but must be top-level or static.

**Additional context**
I'm working on a package that works with a cms REST API so, structures are dynamic. Devs can create their custom classes, this new funcitions will help to deserialize the data.

This example from JsonSerializable can be a better reference :smile:
https://github.com/google/json_serializable.dart/blob/master/example/lib/tuple_example.dart

fabianMendez:
Hi, this is currently a work in progress as I have yet to refactor some repeated code and probably add some tests and documentation.

Fixes #331 

## What works:
- Multiple generic types: It generates the name of the function based on the type's name, e.g: fromJsonT, fromJsonS, etc...
- Named parameters (marked with required)
- Named constructors: It always requires all the fromJson and toJson functions

## What does not work:

- Nullable generic types: Json Serializable always expects the value returned by the fromJsonT/toJsonT functions to be not null. See [this](https://github.com/google/json_serializable.dart/issues/803)


packages/freezed/lib/src/templates/tear_off.dart
```diff
@@ -19,6 +19,11 @@ class TearOff {
   final GenericsDefinitionTemplate genericsDefinition;
   final List<ConstructorDetails> allConstructors;
 
+  bool get _hasGenericArgumentFactories =>
+      allConstructors.any((cons) => cons.decorators
+          .where((dec) => dec.startsWith('@JsonSerializable'))
+          .any((dec) => dec.contains('genericArgumentFactories: true')));
```
rrousselGit, line 25: We can't rely on the "toString" of the annotation

It should be possible to do:

```dart
const alias = JsonSerializable(genericArgumentFactories: true);

...
@alias
factory MyClass() = _MyClass;
```

fabianMendez: Yep, I was thinking on doing something like this:

```dart
  bool get hasGenericArgumentFactories {
    if (!hasJsonSerializable) return false;
    final annotation =
        const TypeChecker.fromRuntime(JsonSerializable).firstAnnotationOf(this);
    return annotation.getField('genericArgumentFactories').toBoolValue();
  }
```

But I have not been able to add a field to the `ConstructorDetails` class

fabianMendez:
Hi @rrousselGit I think this is ready, what do you think?

packages/freezed/README.md
```diff
@@ -1229,6 +1230,44 @@ class MyModel with _$MyModel {
...
+  factory GenericWithArgumentFactories.fromJson(
+          Map<String, dynamic> json, T Function(Object? json) fromJsonT) =>
+      _$GenericWithArgumentFactoriesFromJson<T>(json, fromJsonT);
+}
+```
+
+Which can be used as follows:
+
+```dart
+GenericWithArgumentFactories.fromJson(
```
rrousselGit, line 1254:
```suggestion
GenericWithArgumentFactories<GenericValue>.fromJson(
```

To make it clearer what that `GenericValue` is

packages/freezed/README.md
```diff
@@ -1229,6 +1230,44 @@ class MyModel with _$MyModel {
 }
 ```
 
+### classes with generic argument factories
+
+If you annotate your constructor(s) with `@JsonSerializable(genericArgumentFactories: true)`, [Freezed] will
+require a `fromJson` and `toJson` function for each generic type. For example, the following class requires an aditional `fromJsonT` and `toJsonT`:
```
rrousselGit, line 1236: I don't see an example for the `toJsonT`

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -186,6 +189,11 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
     return false;
   }
 
+  bool _hasGenericArgumentFactories(ClassElement element) {
```
rrousselGit, line 192: For union types, what if one constructor uses `hasGenericArgument` but another doesn't?

We probably should check this case and throw an error

rrousselGit: Similarly, we probably want a:

```dart
@Freezed(hasGenericArgument: true)
```

to avoid having to pass a `@JsonSerializable` on all the constructors of a union

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -545,6 +557,15 @@ extension on Element {
       throwOnUnresolved: false,
     );
   }
+
+  bool get hasGenericArgumentFactories {
+    if (!hasJsonSerializable) return false;
+
+    final annotation =
+        const TypeChecker.fromRuntime(JsonSerializable).firstAnnotationOf(this);
+    final value = annotation.getField('genericArgumentFactories').toBoolValue();
```
rrousselGit, line 566: missing null check? firstAnnotationOf can return null

packages/freezed/lib/src/freezed_generator.dart
fabianMendez: If the other constructor does not use the `hasGenericArgument` annotation and has a generic argument, json_serializable throws an error.

packages/freezed/lib/src/templates/abstract_template.dart
```diff
@@ -51,7 +53,13 @@ ${copyWith.commonContreteImpl(abstractProperties)}
 
   String get _toJson {
     if (!shouldGenerateJson) return '';
-    return 'Map<String, dynamic> toJson() => throw $privConstUsedErrorVarName;';
+    final genericToJsonArgs = hasGenericArgumentFactories
+        ? genericsParameter.typeParameters.map((type) {
+            return 'Object Function($type value) toJson$type';
```
rrousselGit, line 58:
```suggestion
            return 'Object? Function($type value) toJson$type';
```

We also need a test to check that toJson can return null

packages/freezed/lib/src/freezed_generator.dart
fabianMendez: The line before checks if it does not have the `JsonSerializable` annotation and returns false. Should we still check if `firstAnnotationOf` returns null?

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -154,12 +156,32 @@ ${copyWith.abstractCopyWithGetter}
...
 
   String get _concreteFromJsonConstructor {
     if (!shouldGenerateJson) return '';
-    return 'factory $concreteName.fromJson(Map<String, dynamic> json) => _\$${concreteName}FromJson(json);';
+
+    final genericArgs = hasGenericArgumentFactories
+        ? genericsParameter.typeParameters.map((type) {
+            return ', $type Function(Object? json) fromJson$type';
+          }).join()
+        : '';
```
rrousselGit, line 176: to extract in a function to reuse with _redirectedFromJsonConstructor

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -154,12 +156,32 @@ ${copyWith.abstractCopyWithGetter}
...
+
+    final genericArgsNames = hasGenericArgumentFactories
+        ? genericsParameter.typeParameters
+            .map((type) => ', fromJson$type')
+            .join()
+        : '';
```
rrousselGit, line 182: same thing, make a reusable function

rrousselGit: In fact we probably should extract all the toJson logic in a separate template file

packages/freezed/test/json_test.dart
```diff
@@ -133,6 +133,130 @@ Future<void> main() async {
         Generic(42),
       );
     });
+
+    group('with argument factories', () {
```
rrousselGit, line 137: Make sure to specify the generic types inside tests instead of relying on type inference

This ensures that fromJson wouldn't be deserializing to Object/dynamic somehow

packages/freezed/lib/src/templates/abstract_template.dart
fabianMendez: `JsonSerializable` does not currently support nullable types, for instance if we accept a nullable type for the `toJson` functions, it complains:
imagen

packages/freezed/lib/src/freezed_generator.dart
rrousselGit: it may not if we rely on TypeConverter, and we probably shouldn't rely on json_serializable for this anyway. Freezed should have its own error handling

packages/freezed/lib/src/templates/abstract_template.dart
rrousselGit: Hum that's problematic

packages/freezed/lib/src/templates/concrete_template.dart
fabianMendez: I've extracted it into a new private getter

packages/freezed/lib/src/templates/concrete_template.dart
fabianMendez: I don't see where this is repeated

packages/freezed/README.md
fabianMendez: I've separated the fromJson and toJson examples

packages/freezed/lib/src/freezed_generator.dart
fabianMendez: I am sorry, did you mean `TypeChecker`?, I am not familiar with `TypeConverter`

packages/freezed/test/json_test.dart
fabianMendez: Done

packages/freezed/lib/src/templates/concrete_template.dart
rrousselGit: It's repeated accross files.

As I said before, we'd want to extract this logic in a separate `templates/serialisation_template.dart` with all the fromJson/toJson logic.

packages/freezed/lib/src/freezed_generator.dart
rrousselGit: TypeConverted is from JsonSerializable

We still need the `@Freezed(hasGenericArgument: true)` annotation and the exception

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -545,6 +557,15 @@ extension on Element {
       throwOnUnresolved: false,
     );
   }
+
+  bool get hasGenericArgumentFactories {
+    if (!hasJsonSerializable) return false;
```
rrousselGit, line 562:
```suggestion
```

packages/freezed/lib/src/freezed_generator.dart
rrousselGit, line 566:
```suggestion
    final value = annotation?.getField('genericArgumentFactories').toBoolValue();
```

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -545,6 +557,14 @@ extension on Element {
       throwOnUnresolved: false,
     );
   }
+
+  bool get hasGenericArgumentFactories {
+
+    final annotation =
+        const TypeChecker.fromRuntime(JsonSerializable).firstAnnotationOf(this);
```
rrousselGit, line 564:
```suggestion
    final annotation =
        const TypeChecker.fromRuntime(JsonSerializable).firstAnnotationOf(this);

```

rrousselGit (changes requested):
There are still some unresolved comments :smile:

packages/freezed/lib/src/templates/concrete_template.dart
fabianMendez: Done. Now all fromJson and toJson related logic is inside of `templates/serialization_template.dart`

packages/freezed/lib/src/freezed_generator.dart
fabianMendez: Oh, I think you mean `JsonConverter`?

But still I'm failing to see what the use case could be, because something like this would not require argument factories for the generic types:

```dart
@Freezed(genericArgumentFactories: true)
class Generic<T> with _$Generic<T> {
  factory Generic(@DataConverter() T a) = _Generic<T>;

  factory Generic.fromJson(Map<String, dynamic> json) =>
      _$GenericFromJson<T>(json);
}

class DataConverter<T> implements JsonConverter<T, Object?> {
  const DataConverter();

  @override
  T fromJson(Object? json) {
    return json as T;
  }

  @override
  Object? toJson(T object) {
    return object;
  }
}
```

The only use case I can think of is when using the `generic_argument_factories` option for `json_serializable` in the `buld.yaml` file. In such case we can not rely in the `@JsonSerializable` annotation so it would be required to hint freezed using `@Freezed(genericArgumentFactories: true)` at the class level.

sooxt98:
@fabianMendez I just tried with your latest commit, and i get this issue
image
image

fabianMendez:
> @fabianMendez I just tried with your latest commit, and i get this issue
> 
> image
> 
> image


You should convert that fromJson factory to an expression body:
```dart
@freezed
class ApiResponse<T> with _$ApiResponse<T> {
  @JsonSerializable(genericArgumentFactories: true)
  factory ApiResponse(
    String? status,
    T? data,
    String? message,
  ) = _ApiResponse<T>;

  factory ApiResponse.fromJson(
    Map<String, dynamic> json,
    T Function(Object? json) fromJsonT,
  ) => _$ApiResponseFromJson<T>(json, fromJsonT);
}
```

Dunno why, but that's how it is. I mean this is a change I did not introduce (I hope)...

sooxt98:
@fabianMendez yay it works, my bad for not using the arrow func. But somehow this patch bricked retrofit;

image

It should be like this, the param is missing
image

sooxt98:
@fabianMendez I solved it by upgrading source_gen to latest version that support null-safety lol

packages/freezed/lib/src/freezed_generator.dart
rrousselGit: It's not really about a use-case. It's about not having Freezed depend on how json_serializable behaves.

We need Freezed to provide a custom error message

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -424,14 +432,18 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
     GlobalData globalData,
     Data data,
   ) sync* {
+    final commonSerialization = Serialization(
```
rrousselGit, line 435: Neat

rrousselGit:
Could you fix conflicts? (sorry for that)

Other than that, it should be ready to go :smile:

sooxt98:
@rrousselGit any idea with my issue? This commit is breaking others package shouldnt be commit without a fix

rrousselGit:
@sooxt98 I don't understand what your issue is. What does "cause it doesn't generate Model.fromJson for Retrofit" mean? As there's no `Model` in your snippet.

sooxt98:
@rrousselGit please refer to the comment with two image comparison; that's the missing piece that should generate by freezed; line 33; 

The comment after it is the code use to generate those two pictures

packages/freezed/lib/src/freezed_generator.dart
fabianMendez: I've added an assertion to verify that if one constructor is annotated with `@JsonSerializable(genericArgumentFactories: true)` all other constructors are annotated as well.

https://github.com/fabianMendez/freezed/blob/9a9fb9d658af4f0892b709825ebff76236c99784/packages/freezed/lib/src/freezed_generator.dart#L158-L178

What do you think?

ibrahim-mubarak:
This doesn't seem to work when `generic_argument_factories: true` has been set on `build.yaml`
Each generic class will have to be specifically annotated.

rrousselGit:
As a comeback to this:

I have detected some issues (including json_serializable throwing a RangeError in some conditions). I'm trying to pinpoint what the problem is and raise a PR

rrousselGit:
It appears that the RangeError issue is caused by such objects:

```dart
@freezed
class Either<First, Second> with _$Either<First, Second> {
  factory Either.first(First value) = _EitherFirst;
  factory Either.second(Second value) = _EitherSecond;

  factory Either.fromJson(
    Map<String, Object?> json,
    First Function(Object?) fromJsonFirst,
    Second Function(Object?) fromJsonSecond,
  ) =>
      _$EitherFromJson(json, fromJsonFirst, fromJsonSecond);
}

@freezed
class Another with _$Another {
  factory Another(Either<int, String> value) = _Another;

  factory Another.fromJson(Map<String, Object?> json) =>
      _$AnotherFromJson(json);
}
```

The problem being, `toJson` is defined by the mixin `_$Either`. But the class annotated with `@JsonSerializable` is another class

So `json_serializable` gets confused and is unable to map the generic parameters of the mixins with the generic parameters of the class.

rrousselGit:
One solution appears to be to, by default, generate a `toJson` with no argument.

So instead of:

```dart
  Map<String, dynamic> toJson(
    Object Function(First) toJsonFirst,
    Object Function(Second) toJsonSecond,
  );
```

we do:

```dart
  Map<String, dynamic> toJson();
```

This appears to be compatible with `genericArgumentFactories`.

And if we want to support those arguments, the syntax may be:

```dart
@freezed
class Either<First, Second> with _$Either<First, Second> {
  factory Either.first(First value) = _EitherFirst;
  factory Either.second(Second value) = _EitherSecond;

  factory Either.fromJson(
    Map<String, Object?> json,

    First Function(Object?) fromJsonFirst,
    Second Function(Object?) fromJsonSecond,
  ) =>
      _$EitherFromJson(json, fromJsonFirst, fromJsonSecond);

  // override the toJson so that json_serializable is able to understand how the
  // generic parameters between Either and _$Either are related
  @override
  Map<String, dynamic> toJson(
    Object Function(First) toJsonFirst,
    Object Function(Second) toJsonSecond,
  );
}
```



Alternatively, a PR could be made on json_serializable to support such a case.

fabianMendez:
Hi @rrousselGit I am so sorry for practically abandoning this PR, I got lost in the advanced use/edge cases. Would it be possible to merge what is working now and let users test and report specific use cases so that they can be tackle individually?

PR #372, opened 2021-03-02
Fix `sort_unnamed_constructors_first` and `cast_nullable_to_non_nullable`

gaetschwartz:
This fixes 2 lints. 

The first one being `sort_unnamed_constructors_first`, the abstract template didn't have the unnamed constructor so if you had the lint enabled it would complain. 

Second one is related to null safety, it doesn't like when we cast potentially nullable objects (`Object?`) as OneClass because it makes it non nullable and it's not a good practice. I added a `!` when the class is non-nullable. 

It passes all tests.

### DIFF
--- a/packages/freezed/lib/src/templates/copy_with.dart
+++ b/packages/freezed/lib/src/templates/copy_with.dart
@@ -177,11 +177,14 @@ $_abstractClassName${genericsParameter.append('$clonedClassName$genericsParamete
     @required String returnType,
   }) {
     String parameterToValue(Parameter p) {
-      var ternary = '${p.name} == freezed ? $accessor.${p.name} : ${p.name}';
-      if (p.type != 'Object?' && p.type != null) {
-        ternary = '$ternary as ${p.type}';
+      final ternary = '${p.name} == freezed ? $accessor.${p.name} : ${p.name}';
+      var cast = '';
+      if (p.type != 'Object?' && p.type != 'Object' && p.type != null) {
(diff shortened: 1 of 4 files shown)


packages/freezed/lib/src/templates/copy_with.dart
```diff
@@ -177,11 +177,15 @@ $_abstractClassName${genericsParameter.append('$clonedClassName$genericsParamete
     @required String returnType,
   }) {
     String parameterToValue(Parameter p) {
-      var ternary = '${p.name} == freezed ? $accessor.${p.name} : ${p.name}';
-      if (p.type != 'Object?' && p.type != null) {
-        ternary = '$ternary as ${p.type}';
+      final ternary = '${p.name} == freezed ? $accessor.${p.name} : ${p.name}';
+      var cast = '';
+      if (p.type != null) {
+        cast = p.type.contains('?') ? '' : '!';
```
rrousselGit, line 183: I'd locally ignore the nnbd lint instead

The '!' operator costs an operation at runtime, which we don't want

gaetschwartz: Alright, figured you were not very fond of ignoring lints, will do.

rrousselGit: Sorry I meant "ignore this line" not "ignore for file"

The lint could be useful in other places in this file, just not here

rrousselGit:
Perfect thanks!

PR #381, opened 2021-03-06
Add union value customization

### LINKED ISSUE #360: Ability to declare union key value for factory constructor
Not sure if this already exists or not but from my experience the unionKey value for a union class constructor is determined by the part that comes after the base class name  (i.e. `SomeClass.fooBar(...)` would produce a union key value of `"fooBar"`. 

Is it possible to override the value for the union key? I've done this in the past by naming the constructor to match what I need it to output as i.e. `SomeClass.foo_bar(...)` to produce `foo_bar` but kind of a hack and doesn't allow for special characters. 

Essentially what I'm looking for is some annotation that would allow me to declare what the unionKey is for each factory constructor explicitly

Thank you for all of the effort you put into this awesome package don't know what I'd do without it!

ookami-kb:
This is a draft PR to discuss the approach. If it's ok, I guess we have to break it into 2 PRs and publish a new version of `freezed_annotation` with `FreezedUnionValue` first.

It should resolve #360

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -1141,12 +1141,19 @@ Then [Freezed] will use each JSON object's `runtimeType` to choose the construct
 ]
 ```
 
-You can customize this key to replace `runtimeType` with something different
-using the `@Freezed` decorator:
+You can customize key and value with something different
+using `@Freezed` and `@FreezedUnionValue` decorators:
 
 ```dart
 @Freezed(unionKey: 'type')
 abstract class MyResponse with _$MyResponse {
+  const factory MyResponse(String a) = MyResponseData;
+  
+  @FreezedUnionValue('SpecialCase')
+  const factory MyResponse.special(String a, int b) = MyResponseSpecial;
+  
+  const factory MyResponse.error(String message) = MyResponseError;
+
   // ...
 }
 ```
@@ -1160,7 +1167,7 @@ which would update the previous json to:
     "a": "This JSON object will use constructor MyResponse()"
   },
   {
-    "type": "special",
+    "type": "SpecialCase",
     "a": "This JSON object will use constructor MyResponse.special()",
     "b": 42
   },
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -1145,7 +1145,7 @@ You can customize key and value with something different
 using `@Freezed` and `@FreezedUnionValue` decorators:
 
 ```dart
-@Freezed(unionKey: 'type')
+@Freezed(unionKey: 'type', unionValueCase: FreezedUnionCase.pascal)
 abstract class MyResponse with _$MyResponse {
   const factory MyResponse(String a) = MyResponseData;
   
@@ -1163,7 +1163,7 @@ which would update the previous json to:
 ```json
 [
   {
-    "type": "default",
+    "type": "Default",
     "a": "This JSON object will use constructor MyResponse()"
   },
   {
@@ -1172,7 +1172,7 @@ which would update the previous json to:
     "b": 42
   },
   {
-    "type": "error",
+    "type": "Error",
     "message": "This JSON object will use constructor MyResponse.error()"
   }
 ]
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -1178,6 +1178,19 @@ which would update the previous json to:
 ]
 ```
 
+If you want to customize key and value for all the classes, you can specify it inside your
+`build.yaml` file, for example:
+
+```yaml
+targets:
+  $default:
+    builders:
+      freezed:
+        options:
+          union_key: type
+          union_value_case: pascal
+```
+
 If you don't control the JSON response, then you can implement a custom converter.
 Your custom converter will need to implement its own logic for determining which
 constructor to use.
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -550,11 +550,12 @@ extension on Element {
 
 extension on ConstructorElement {
   String get unionValue {
-    return const TypeChecker.fromRuntime(FreezedUnionValue)
-            .firstAnnotationOf(this, throwOnUnresolved: false)
-            ?.getField('value')
-            ?.toStringValue() ??
-        (isDefaultConstructor(this) ? 'default' : name);
+    final annotation = const TypeChecker.fromRuntime(FreezedUnionValue)
+        .firstAnnotationOf(this, throwOnUnresolved: false);
+    if (annotation != null) {
+      return annotation.getField('value').toStringValue();
+    }
+    return isDefaultConstructor(this) ? 'default' : name;
   }
 }
 
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -580,9 +580,9 @@ extension on ConstructorElement {
         return pascalCase(constructorName);
       case FreezedUnionCase.snake:
         return snakeCase(constructorName);
+      default:
+        throw FallThroughError();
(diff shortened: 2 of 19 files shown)


packages/freezed/pubspec.yaml
```diff
@@ -14,7 +15,8 @@ dependencies:
   build_config: ^0.4.2
   meta: ^1.3.0
   source_gen: ^0.9.0
-  freezed_annotation: ">=0.14.0 <2.0.0"
+  freezed_annotation:
```
rrousselGit, line 18: I'd suggest using `melos` instead, with the command `melos boostrap`

packages/freezed_annotation/lib/freezed_annotation.dart
```diff
@@ -200,3 +200,9 @@ class With {
   final Type? type;
   final String? stringType;
 }
+
+class FreezedUnionValue {
```
rrousselGit, line 204: We'll need this commented and documented in the readme

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -547,6 +548,16 @@ extension on Element {
   }
 }
 
+extension on ConstructorElement {
+  String get unionValue {
+    return const TypeChecker.fromRuntime(FreezedUnionValue)
+            .firstAnnotationOf(this, throwOnUnresolved: false)
+            ?.getField('value')
+            ?.toStringValue() ??
+        (isDefaultConstructor(this) ? 'default' : name);
```
rrousselGit, line 557: Could you remove the `??` and split this in multiple lines? For readability

rrousselGit:
It's looking good, thanks for your help!

rrousselGit:
What do you think about adding an option on `@Freezed(...)` to alternative use the constructor name, but with a case change

Like:


```dart
@Freezed(unionCase: FreezedUnionCase.snakeCase)
``` 

Which for `MyClass.myName` would use `my_name` as union key

packages/freezed/pubspec.yaml
ookami-kb: Cool, didn't know about that, will take a look.

packages/freezed_annotation/lib/freezed_annotation.dart
ookami-kb: Sure, I just wanted to make sure if you're ok with this approach first.

ookami-kb:
> What do you think about adding an option on `@Freezed(...)` to alternatively use the constructor name, but with a case change
> 
> Like:
> 
> ```dart
> @Freezed(unionCase: FreezedUnionCase.snakeCase)
> ```
> 
> Which for `MyClass.myName` would use `my_name` as union key

Yeah, I think it could be useful as well. I guess I will follow the same way as they do it in `json_serializable`: https://github.com/google/json_serializable.dart/blob/6b5cfd6087e38308988b635392a558154a923613/json_serializable/lib/src/utils.dart#L30

ookami-kb:
It should be ready now.

packages/freezed/README.md
```diff
@@ -1141,12 +1141,19 @@ Then [Freezed] will use each JSON object's `runtimeType` to choose the construct
...
 ```
 
-You can customize this key to replace `runtimeType` with something different
-using the `@Freezed` decorator:
+You can customize key and value with something different
+using `@Freezed` and `@FreezedUnionValue` decorators:
 
 ```dart
-@Freezed(unionKey: 'type')
+@Freezed(unionKey: 'type', unionValueCase: FreezedUnionCase.pascal)
```
rrousselGit, line 1148: that's neat!

packages/freezed_annotation/lib/freezed_annotation.dart
```diff
@@ -57,6 +60,38 @@ class Freezed {
...
+  /// {@macro freezed_annotation.freezed.example}
+  ///
+  /// This variable allows customizing the value used (constructor name by
+  /// default).
+  ///
+  /// For example, we could change our previous `Union` implementation to:
+  ///
+  /// ```dart
+  /// @Freezed(unionValueCase: FreezedUnionCase.pascal)
+  /// abstract class Union with _$Union {
```
rrousselGit, line 78:
```suggestion
  /// class Union with _$Union {
```

packages/freezed_annotation/lib/freezed_annotation.dart
```diff
@@ -9,12 +9,14 @@ class Freezed {
...
   });
 
   /// Determines what key should be used to de/serialize union types.
   ///
   /// Consider:
   ///
+  /// {@template freezed_annotation.freezed.example}
   /// ```dart
   /// @freezed
   /// abstract class Union with _$Union {
```
rrousselGit, line 22:
```suggestion
  /// class Union with _$Union {
```

packages/freezed/test/json_test.dart
```diff
@@ -88,6 +88,38 @@ Future<void> main() async {
     });
   });
 
+  group('FreezedUnionValue', () {
```
rrousselGit, line 91: Could you add tests for all the different "case"?

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -547,6 +562,30 @@ extension on Element {
...
+        return constructorName;
+      case FreezedUnionCase.kebab:
+        return kebabCase(constructorName);
+      case FreezedUnionCase.pascal:
+        return pascalCase(constructorName);
+      case FreezedUnionCase.snake:
+        return snakeCase(constructorName);
+    }
+
+    throw ArgumentError('Unknown unionCase: $unionCase}');
```
rrousselGit, line 585: Instead do:

```dart
case ...:
default:
  throw FallthroughError();
```

rrousselGit:
Awesome! Just needs a few tests for the different name cases, and it'll be good to merge :smile:

packages/freezed/lib/src/freezed_generator.dart
ookami-kb: I'm ok to change it, but personally I don't like `default` in enum switch, because without it, the analyser will at least give a warning if another case is added to enum.

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -348,8 +351,20 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
         configs['union_key']?.toString() ??
         'runtimeType';
 
+    FreezedUnionCase unionValueCase;
+    final fromConfig = configs['union_value_case']?.toString();
+    if (fromConfig != null) {
+      unionValueCase = FreezedUnionCase.values.firstWhere((e) =>
+          e.toString().replaceFirst('FreezedUnionCase.', '') == fromConfig);
+    } else {
```
rrousselGit, line 359: I see that you added a way to specify this in the `build.yaml`!

That's cool. Could you document it in the READMe and test add an e2e test for it? (you can update the config of `example` and add the test in `example/test`)

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -348,8 +351,20 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
         configs['union_key']?.toString() ??
         'runtimeType';
 
+    FreezedUnionCase unionValueCase;
+    final fromConfig = configs['union_value_case']?.toString();
+    if (fromConfig != null) {
+      unionValueCase = FreezedUnionCase.values.firstWhere((e) =>
```
rrousselGit, line 357: I'd prefer a swtch-case over the `replaceFirst`

rrousselGit:
That's awesome!

I think this is good to go. I'll merge it. Thanks for your help!

PR #382, opened 2021-03-08
Readme fix pre-null version

jlnrrg:
freezed_annotation has no version 0.12.7 published, thus 0.12.0 is required
Also I added freezed for clarification.


rrousselGit:
Oopsy, thanks!

PR #388, opened 2021-03-12
fix: Take nested exports into account

ookami-kb:
Our current setup: we have a package `core` with the following content:

```dart
library core;

export 'package:freezed_annotation/freezed_annotation.dart';
// ... other exports and imports
```

Also, we have a dependent package where we use freezed with something like this:

```dart
import 'package:core/core.dart';

part 'get_services.freezed.dart';

part 'get_services.g.dart';

@freezed
class GetServicesResponseDto with _$GetServicesResponseDto {
  const factory GetServicesResponseDto({
    required List<ServiceDto> services,
  }) = _GetServicesResponseDto;

  factory GetServicesResponseDto.fromJson(Map<String, dynamic> json) => _$GetServicesResponseDtoFromJson(json);
}
```

The problem is that `get_services.g.dart` is not generated, but it is generated if I add directly `import 'package:freezed_annotation/freezed_annotation.dart';`

The proposed fix solves the problem, since `export.isRelevant(whereLibrary)` blocks locator from going deeper into exported libraries. But maybe I'm missing something and this code is needed for something else?

### DIFF
--- a/packages/freezed/lib/src/tools/recursive_import_locator.dart
+++ b/packages/freezed/lib/src/tools/recursive_import_locator.dart
@@ -52,12 +52,11 @@ class RecursiveImportLocator {
     }
 
     return library.exportedLibraries.any((export) {
-      return export.isRelevant(whereLibrary) &&
-          _doesExportRecursively(
-            library: export,
-            where: where,
-            whereLibrary: whereLibrary,
-          );
+      return _doesExportRecursively(
+        library: export,
+        where: where,
+        whereLibrary: whereLibrary,
+      );
     });
   }
 }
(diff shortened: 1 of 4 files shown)


rrousselGit (changes requested):
Thanks,

Could you add tests for this?

ookami-kb:
No idea how to cover this with tests, but I will try to come up with something.

rrousselGit:
make a file inside `integration` that exposes `freezed_annotation`, then make a freezed class that uses this import (placed in integration/export.dart)

Then add a test that the class works as expected

rrousselGit:
Perfect thanks!

PR #390, opened 2021-03-13
Instructions for the projects that use non-nullable feature.

yfer:
I have detailed report of what is the problem [here](https://github.com/dart-lang/build/issues/3046). and [here](https://github.com/dart-lang/build/pull/3047)
Basically Freezed package is not usable when you use non-nullable feature in you project (for example after migration to sound null-safety). 
It seems that package will not be usable without these instructions until build package will be null-safe. I don't think they will use this trick on the package itself. But information should be somewhere for us users.

By the way. Thank you for your great project!


rrousselGit:
This is false, Freezed is null-safety ready.


Make sure to upgrade your Freezed version to the latest available (0.14.0)

yfer:
@rrousselGit 

Freezed itself is properly null-safe.
I am using 0.14.0+1, and problem definitely not with the package itself.
Problem is with tooling, something seems broken with flutter|dart tooling and build_runner package. Maybe it was not broken in some earlier version but it is on.
```
    • Flutter version 2.0.2 at C:\flutter
    • Framework revision 8962f6dc68 (3 days ago), 2021-03-11 13:22:20 -0800
    • Engine revision 5d8bf811b3
    • Dart version 2.12.1
```

I will track what will happen with my Issue there on the tracker and write results here.

dancamdev:
Any update on this? `build_runner` not being null safe is kinda breaking every other package depending on it

rrousselGit:
@dancamdev Could you clarify?

dancamdev:
Yeah, I've put it in a bad way, I mean that `build_runner` not being null-safe will make every project depending on it in some way also not sound-null-safe, am I right? Therefore, since `build_runner` is used so broadly, I feel its migration should be prioritized in some way.

yfer:
@dancamdev 
It seems they've incorporated some workaround in earlier versions, but it is not working for now.(at least for me and i've digged it for hours)
So the fastest way for now is just to add // @dart=2.9 at the beginning of flutter/.pub-cache/hosted/pub.dartlang.org/build_runner-1.12.1/bin/build_runner.dart

rrousselGit:
> I mean that build_runner not being null-safe will make every project depending on it in some way also not sound-null-safe, am I right? 

No that is not the case

`build_runner` is used by terminals/IDEs, not Flutter apps. It not being migrated has no influence.

What matters is that code-generators generate null-safe code, which Freezed does.

rrousselGit:
Closing as https://github.com/rrousselGit/freezed/pull/404 should take care of the issue (build_runner is now migrated)

PR #422, opened 2021-04-11
Correct README example

creativecreatorormaybenot:
The previous version does not build/compile because:

> Final variables require a MyClass._() constructor

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -409,7 +409,7 @@ for computed properties.
 For example, you may write:
 
 ```dart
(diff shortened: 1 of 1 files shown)


packages/freezed/README.md
```diff
@@ -410,6 +410,8 @@ For example, you may write:
 
 ```dart
 abstract class Todos with _$Todos {
```
rrousselGit, line 412:
```suggestion 
class Todos with _$Todos {
```

rrousselGit:
Thanks!

PR #426, opened 2021-04-13
being explicit about json_serializable dependency

lesnitsky:
I keep forgetting to do this and always come back to the doc to see what went wrong and find myself doing everything right, but still not getting `.g.dart` in my file system. I guess this note will help next time


rrousselGit:
Fair enough

PR #430, opened 2021-04-20
Feature/union fallback

Brazol:
Hi, this is a small change to support fallback option for union types. Currently when type is not matched by factory constructor name or using custom **FreezedUnionValue** annotation generated code will throw **FallThroughError**. I've added **FreezedUnionFallback** annotation to mark constructor that should be used as a fallback one.

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -285,6 +285,13 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
       );
     }
 
+    if (configs.fallbackUnion != null && result.none((c) => c.isFallback)) {
+      throw InvalidGenerationSourceError(
+        'Fallback union was specified but no ${configs.fallbackUnion} constructor exists.',
+        element: element,
+      );
+    }
+
     return result;
   }
 
@@ -371,9 +378,7 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
         configs['union_key']?.toString() ??
         'runtimeType';
 
-    final fallbackUnion =
-        annotation.getField('fallbackUnion')?.toStringValue() ??
-            configs['fallback_union']?.toString();
+    final fallbackUnion = annotation.getField('fallbackUnion')?.toStringValue();
 
     FreezedUnionCase unionValueCase;
     final fromConfig = configs['union_value_case']?.toString();
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -255,7 +255,7 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
           implementsDecorators:
               _implementsDecorationTypes(constructor).toSet().toList(),
           isDefault: isDefaultConstructor(constructor),
-          isFallback: constructor.isUnionFallbackConstructor,
+          isFallback: constructor.isFallbackUnion(configs.fallbackUnion),
           hasJsonSerializable: constructor.hasJsonSerializable,
           cloneableProperties: await _cloneableProperties(
             buildStep,
@@ -371,6 +371,9 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
         configs['union_key']?.toString() ??
         'runtimeType';
 
+    final fallbackUnion = annotation.getField('fallbackUnion')?.toStringValue() ??
+        configs['fallback_union']?.toString();
+
     FreezedUnionCase unionValueCase;
     final fromConfig = configs['union_value_case']?.toString();
     switch (fromConfig) {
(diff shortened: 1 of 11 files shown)


rrousselGit:
Thanks!

One quick thing: Rather than a new annotation, we probably could do:

```dart
@Freezed(fallbackUnion: 'constructorName' | 'default')
```

Brazol:
> Thanks!
> 
> One quick thing: Rather than a new annotation, we probably could do:
> 
> ```dart
> @Freezed(fallbackUnion: 'constructorName' | 'default')
> ```

Sure. Changed to use Freezed annotation instead of new one.

Brazol:
@rrousselGit any chance of getting this merged? Or is there something more to do here? :)

rrousselGit:
Oh sorry I didn't get the notification!
I will take a look today.

packages/freezed/test/json_test.dart
```diff
@@ -120,6 +120,61 @@ Future<void> main() async {
     });
   });
 
+  group('FreezedUnionFallback', () {
+    test('fromJson', () {
```
rrousselGit, line 124: Could you add a test for when there's no `runtimeType` key at all?

packages/freezed/lib/src/templates/from_json_template.dart
```diff
@@ -35,11 +37,19 @@ class FromJson {
         ''';
       }).join();
 
+      var defaultCase = 'throw FallThroughError();';
+      if (constructors.any((element) => element.isFallback)) {
```
rrousselGit, line 41: Rather than `any`, do the `singleWhere` query before the if, and do `fallbackConstructor != null`. To avoid iterating over the list twice.

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -370,6 +371,10 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
         configs['union_key']?.toString() ??
         'runtimeType';
 
+    final fallbackUnion =
+        annotation.getField('fallbackUnion')?.toStringValue() ??
+            configs['fallback_union']?.toString();
```
rrousselGit, line 376: I don't think that specifying a fallback key from the configs makes sense.

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -398,6 +403,7 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
 
     return Freezed(
       unionKey: rawUnionKey.replaceAll("'", r"\'").replaceAll(r'$', r'\$'),
+      fallbackUnion: fallbackUnion,
```
rrousselGit, line 406: We'll need error handling, for cases where the `fallbackUnion ` key is specified but doesn't match a class constructor.

Throwing an `InvalidGenerationSourceError` with the error message would make sense.
It'll need a test, which you can test inside `test/source_gen_test.dart`/`source_gen_src.dart`

packages/freezed/lib/src/templates/from_json_template.dart
```diff
@@ -35,11 +37,19 @@ class FromJson {
...
+        final fallbackConstructor =
+            constructors.singleWhere((element) => element.isFallback);
+        defaultCase =
+            'return ${fallbackConstructor.redirectedName}$genericParameters.fromJson(json);';
```
rrousselGit, line 45:
```suggestion
  // TODO(rrousselGit): update logic once https://github.com/rrousselGit/freezed/pull/370 lands
            'return ${fallbackConstructor.redirectedName}$genericParameters.fromJson(json);';
```

rrousselGit:
Thanks, it's looking pretty good!

I've made a few comments, after which is should be good to go!

packages/freezed/lib/src/freezed_generator.dart
Brazol: I thought about the case when someone wants to just use `fallback()` factory everywhere. Or at least in places that factory is created. The same way that unionKey can be globally defined in config. But that would require not throwing error when there is no constructor with specified key (as you pointed out below). So I can remove it.

packages/freezed/test/json_test.dart
Brazol: done

packages/freezed/lib/src/templates/from_json_template.dart
Brazol: done

packages/freezed/lib/src/freezed_generator.dart
Brazol: done

Brazol:
> Thanks, it's looking pretty good!
> 
> I've made a few comments, after which is should be good to go!

@rrousselGit thanks for looking into it. I made changes you asked for. :)

rrousselGit (approved):

rrousselGit:
Looking good, thanks!

Issue #226, opened 2020-07-07
Instance of 'UnrecognizedKeysException' - runtimeType


Nolence:
Hi!

Having the `json_serializable` configuration `disallow_unrecognized_keys` set to true throws an error in a class with multiple types as `json_serializable` doesn't know about the field `runtimeType` that is added in each type:

```dart
return _$_$TextModuleToJson(this)..['runtimeType'] = 'textModule';
```

Is there any way to prevent this?

hacker1024:
@rrousselGit

The fix added breaks my code.
The API I'm using has a few types of media.

```dart
enum MediaType {
  @JsonValue('station') station,
  @JsonValue('playlist') playlist,
  @JsonValue('track') track,
}
```

Various objects are related to media types, such as media objects that can be played and media objects that contain metadata.

```dart
abstract class MediaObject {
  MediaType get mediaType;
}

@Freezed(unionKey: 'type')
class PlayableMedia with _$PlayableMedia implements MediaObject {
  @FreezedUnionValue('track')
  const factory PlayableMedia.track({
    @JsonKey(name: 'type') required MediaType mediaType,
    // ...
  });
}

@Freezed(unionKey: 'type')
class MediaDescription with _$MediaDescription implements MediaObject {
  @FreezedUnionValue('track')
  const factory MediaDescription.track({
    @JsonKey(name: 'type') required MediaType mediaType,
    // ...
  });
}
```

It's true that each union type will only ever have its corresponding `mediaType` value, so parsing it from the JSON is a little redundant. That being said, I don't know of any other solution to achieve the same behaviour. As of the latest update, even this method does not work.

hacker1024:
@rrousselGit Gentle bump. Can this change be disable with a configuration option, perhaps?

hacker1024:
@rrousselGit

rrousselGit:
Extracting the JSON key was never officially supported, so no.

If you want this, this probably should be a separate feature request.

Le jeu. 28 avr. 2022 à 09:28, hacker1024 ***@***.***> a
écrit :

Issue #434, opened 2021-04-23
Ability to get all values for union/sealed classes


proninyaroslav:
**Is your feature request related to a problem? Please describe.**

If I need to get all the members of the union/sealed class, then I would like to do the following:
```Dart
final values = [
  Union.one(),
  Union.two(),
  ...
];
```
but this is not without its drawbacks - if a new member is added, then I will have to find for this place in the code, since the analyzer will not work in this case (as if i used `when`/`map`), this creates potential bugs. The appearance of `values` shifts the responsibility of filling the list from the developer to the generator.

**Describe the solution you'd like**

A method/member similar to enum's `values` is needed:
```Dart
final values = Union.values;
```
It may be worth making it optional, so as not to break the API, as is done with `fromJson()`:
```Dart
@freezed
class Union with _$Union {
  const factory Union.one() = UnionOne;
  const factory Union.two() = UnionTwo;
  const factory Union.three() = UnionThree;

  static const values = _$UnionValues;

  factory Union.fromJson(Map<String, dynamic> json) => _$UnionFromJson(json);
}
```

**Additional context**

Yes, it looks like https://github.com/rrousselGit/freezed/issues/74, but the idea is not to create full-fledged enum analogue, but to add `values` analogue, which is very useful in the case of a union/sealed class.

rbdog:
# Problem
I think it becomes hard to implement them when they include properties.

```union.dart
@freezed
class Union with _$Union {
  const factory Union.one( int a ) = UnionOne;
  const factory Union.two( String b ) = UnionTwo;
  const factory Union.three( MyClass c ) = UnionThree;
}
```

# Alternative Solutions

```union.dart
List<Type> derivedClasses = [
  UnionOne, 
  UnionTwo, 
  UnionThree,
];

List<UnionEnum> enumValues = [
  UnionEnum.one,
  UnionEnum.two,
  UnionEnum.three,
];

List<String> names = [
  'one',
  'two',
  'three',
]; 
```

# Other Naming Ideas

- members
- patterns
- subTypes

rrousselGit:
I have no plan to support this. You probably should be using a plain `enum` in this scenario.

Enums can now defined constructors or have methods.

Issue #74, opened 2020-02-27
Support enums


rrousselGit:
Add support for `when`/`map` on custom enums:

```dart
@freezed
abstract class Enum with _$Enum {
  const factory Enum._(String label) = _Enum;

  static const first = Enum._('1st');
  static const second = Enum._('2nd');

  static const values = _$EnumValues;

  factory Enum.fromJson(Map<String, dynamic> json) => _$EnumFromJson(json);
}
```

which then allows

```dart
Enum enum = Enum.first;

enum.when(
  first: (label) {},
  second: (label) {},
);
```

rrousselGit:
This should also make it such that the deserialization does not rely on the contained value either.

Such that changing `static const first = Enum._('1st');` to `static const first = Enum._('first');` won't be a breaking change.

rrousselGit:
Do we need a new decorator for this, or is `@freezed` enough?

I would assume that we don't need one. It could be inferred based on:
- the class has a single private constructor (besides the `toJson`)
- it has a static `values`
- it has at least one `static final`/`static const`  using that class
- it applies only to `static final`/`const`. Static mutable variables are ignored

z-dmitri:
Flags enums are welcome
```dart
Enum enum = Enum.first + Enum.second;
if (enum.has(Enum.first)){
***
}
```

rrousselGit:
That would complexify quite a fair bit the serialization/deserialization process and reduce the readability.

In the end, you can so:

```dart
Set<Enum> enums = {Enum.first, Enum.second};
if (enums.contains(Enum.first)) {

}
```

TarekkMA:
> Do we need a new decorator for this, or is `@freezed` enough?
> 
> I would assume that we don't need one. It could be inferred based on:
> 
> * the class has a single private constructor (besides the `toJson`)
> * it has a static `values`
> * it has at least one `static final`/`static const`  using that class
> * it applies only to `static final`/`const`. Static mutable variables are ignored

Adding a separate annotation like `@freezed_enum` would be better. So user can declare enums explicitly and generator can check for the correct code structure.

And it would be nice if we can also declare a fallback enum value to be used when json input is unknown.

AirborneEagle:
I am certain my question is out of place here, but I can't seem to find an answer elsewhere. (and I don't feel it deserves it's own ticket)

when parsing an enum field, how do I allow the the full string value to come from the Json object.

```
enum Gender {
male, female,
}
...
factory Person(
[...]

PR #456, opened 2021-05-28
Removes union key from json to be deserialized

Nolence:
This addresses #226

I know changes usually require tests but I'm not sure how one could be added. Adding the disallow_unrecognized_keys is in a way enforcing a stricter interface. All Current tests do pass though.

Lemme know


rrousselGit:
Mutating the `json` is not a good idea. This could have unexpected results
Instead consider making a clone

Could you also add a test for this?

Nolence:
I copied every test that involved a unionkey and added annotation specific to jsonserializable.

Added a map copy as well. You're right that this copy is probably less performant but I don't see that many people using the sealed unions to deserialize data anyways. Another option would be to grab the annotation for JsonSerializable and do a check to see if it has the unrecognizedKeys set and then do conditional logic from there

rrousselGit:
What do you think about generating a property with the matching name instead?

Such that we have:

```dart
@JsonSerializable()
class Example {
  @JsonKey()
  final String type;
}
```


This may require changing the default key from `runtimeType` to `type`, which is breaking.
But that's probably a good change anyway

What do you think?

Nolence:
That's a cool idea. Would that require people change their database though or would you be able to

```dart
@JsonSerializable()
class Example {
  @JsonKey(name: 'runtimeType')
  final String type;
}
```
?

Also, instead of string, what about an enum? Not sure of the value of it, but I think it'd possibly be good. We could then generate every enum possibility which would solve #434.

rrousselGit:
> or would you be able to
> 
> ```dart
> @JsonSerializable()
> class Example {
>   @JsonKey(name: 'runtimeType')
>   final String type;
> }
> ```
> 
> ?

I think that would work. But it needs to be tested

We could generate an enum, but I'm not sure if that's worth the trouble. Ultimately for https://github.com/rrousselGit/freezed/issues/74 there's a lot more work to be done than this

Nolence:
> We could generate an enum, but I'm not sure if that's worth the trouble.

No worries, just a random thought I had. Yeah, I'll test out adding that field `type` and wiring things up this weekend and update you with any progress/questions I have

rrousselGit:
Awesome thanks a lot, and sorry for the slow review

rrousselGit:
Hello!

As a comeback to this, what is the reasoning for still removing `type`? Did the `@JsonKey` on a `type` failed to produce the expected result?

PR #460, opened 2021-06-08
Allow a constructor parameter with json converter

### LINKED ISSUE #428: Problem with custom JsonConverter and FieldValue firestore since null safety migration
Hello, 

Well If you have an application using Firestore. Sometimes you may need to use a FieldValue instead of Datetime.
Before null safety, I could run the build runner without any issue with a custom JsonConverter<FieldValue, Object>. 
It actually doesn't work for FieldValue. I did try with JsonConverter<double, Object> and It's working fine.
Does someone has an idea ?

**To Reproduce**

part 'sample.freezed.dart';

part 'sample.g.dart';

@freezed
class Sample with _$Sample {
   const Sample._();
   const factory Sample({
          @ServerTimeStampConverter() double? value1,
          @ServerTimeStampConverter() double? value2,
   }) = _Sample;

    factory Sample.fromJson(Map<String, dynamic> json) =>
      _$SampleFromJson(json);

   factory Sample.fromFirestore(DocumentSnapshot doc) =>
      Sample.fromJson(doc.data()!).copyWith(id: doc.id);
}

ISSUE : 

    ╷
285 │   final FieldValue? updatedOn;
    │                     ^^^^^^^^^
    ╵

**Expected behavior**

Elias8:
Allow constructor parameter with json converter annotation to be generated without making the parameter nullable or using the `@Default()` annotation.

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -132,7 +132,8 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
     for (final parameter in constructor.parameters) {
       if (parameter.type.nullabilitySuffix != NullabilitySuffix.question &&
           parameter.isOptional &&
-          parameter.defaultValue == null) {
+          parameter.defaultValue == null &&
+          !parameter.hasJsonConverter) {
         final ctorName =
             constructor.isDefaultConstructor ? '' : '.${constructor.name}';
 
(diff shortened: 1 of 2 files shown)


rrousselGit:
Hello!
Sorry for the delay.

Could you add tests for this? Thank!

rrousselGit:
This doesn't seem to work, as it generates invalid code for tear-offs.

rrousselGit:
Do you still plan on working on this? Otherwise can I close this?

rrousselGit:
Actually I will close this directly, as thinking about it, this isn't possible at all.

The generate code would never be able to compile due to the `required` keyword

Elias8:
Hi Remi,
Sorry for the inconvenience. 

I was taking a look at it a few days ago, and I found out the issue. 

Is there any workaround you are thinking of?

rrousselGit:
I would suggest making the parameter required

Elias8:
Okay. Thanks!

PR #469, opened 2021-06-30
Allow using enums as FreezedUnionValue values

### LINKED ISSUE #468: Allow using an enum annotated with @JsonValue in the FreezedUnionValue annotation
Using enums as FreezedUnionValue identifiers is safer and easier than using strings.

`json_serializable` supports enum serialization and deserialization, through the use of the `@JsonValue('value')` annotation.

It'd be great to be able to do something like this:

Dart:

```dart
enum MyEnum {
  @JsonValue('value_1')
  value1,

  @JsonValue('value_2')
  value2,
}

@Freezed(unionKey: 'type')
class MyClass with _$MyClass {
  @FreezedUnionValue(MyEnum.value1);
  const factory MyClass.type1(/* ... */) = Type1MyClass;

  @FreezedUnionValue(MyEnum.value2);
  const factory MyClass.type2(/* ... */) = Type2MyClass;
}
```

JSON:

```json
{
  "type": "value_1",
  "...": "..."
}
```

Strings must be used instead.

**Additional context**

hacker1024:
This resolves #468.

There ~~are two~~ is one possible concern~~s~~ with this PR:
1) Integer values, while used as-is when serializing and deserializing enums in json_serializable,
   are converted to strings here. This is because the existing JSON and union code only uses strings.
2) ~~There are no tests. There aren't any exists tests (as far as I can tell) for union annotations at
   all, so I hope this is okay.~~

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -1073,6 +1073,31 @@ which would update the previous json to:
 ]
 ```
 
+An enum can also be used as a union value, optionally using the `JsonValue`
+annotation to override the value string:
+
+```dart
+enum MyResponseType {
+  @JsonValue('Default') defaultResponse,
+  @JsonValue('SpecialCase') specialCase,
+  @JsonValue('Error') error,
+}
+
+@Freezed(unionKey: 'type', unionValueCase: FreezedUnionCase.pascal)
+abstract class MyResponse with _$MyResponse {
+  @FreezedUnionValue(MyResponseType.defaultResponse)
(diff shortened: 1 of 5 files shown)


rrousselGit:
Thanks for your contribution!

As mentioned here https://github.com/rrousselGit/freezed/issues/468#issuecomment-894331915 I will close this to favor a separate solution instead.

Don't take this personally, I simply don't think this is the correct solution to the problem. But I would be happy to solve the problem differently

PR #470, opened 2021-06-30
Throw a CheckedFromJsonException instead of a FallThroughError when an unknown union type value is encountered

hacker1024:
According to the Dart documentation:
> An `Error` object represents a program failure that the programmer
> should have avoided.
>
> Examples include calling a function with invalid arguments,
> or even with the wrong number of arguments,
> or calling it at a time when it is not allowed.
>
> These are not errors that a caller should expect or catch -
> if they occur, the program is erroneous,
> and terminating the program may be the safest response.

In my opinion, a `FallThroughError` is not appropriate to throw when an unknown union type value is deserialized.

- Situations like these may need to be caught, like when receiving arbitrary input in a HTTP request
- `FallThroughError` is an implementation detail - why should us developers care that a switch is being used for this?

This PR replaces the error with a `CheckedFromJsonException`, the exception used by json_serializable when invalid JSON data is deserialised. It allows developers to catch it, and includes more information about its cause.

### DIFF
--- a/packages/freezed/example/lib/main.freezed.dart
+++ b/packages/freezed/example/lib/main.freezed.dart
@@ -175,7 +175,8 @@ Union _$UnionFromJson(Map<String, dynamic> json) {
       return Complex.fromJson(json);
 
     default:
-      throw FallThroughError();
+      throw CheckedFromJsonException(json, 'custom-key', 'Union',
+          'Invalid union type "${json['custom-key']}"!');
   }
 }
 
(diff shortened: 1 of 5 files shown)


hacker1024:
This is not a breaking change; `Error`s are never supposed to be caught.

rrousselGit:
Thanks. Could you add a test for this?

hacker1024:
Sure.

hacker1024:
The main tests seem totally broken; see the GitHub Action log.

rrousselGit:
Those failing tests are caused by the changes introduced.

hacker1024:
Oops, I was misreading the logs. I've added/adapted tests, it should be passing now.

hacker1024:
Actually, it looks like one of the tests is in the wrong group. There's a `CustomUnionValue` test in the `FreezedUnionFallback` group.

PR #480, opened 2021-07-18
Exclude generated files from coverage

### LINKED ISSUE #442: Exclude generated files from coverage

I would like to exclude generated `*.freezed.dart` files when measuring code coverage.

Write `// coverage:ignore-file` to generated files.

Optionally, there could be a flag in `pubspec.yaml` (or `freezed.yaml`?) to make the feature opt-in to avoid surprises for existing users, or if anyone for some reason actually wants to include generated files to coverage.

Even though it's also possible to run `lcov --remove` by hand and (or use [`remove_from_coverage`](https://pub.dev/packages/remove_from_coverage)) remove the undesired files afterwards, it's pretty annoying and clumsy. There are many other code generators out there, of course, but based on a quick Google search, people are often asking about how to exclude generated files, so it seems reasonable to make it easy with Freezed, at least... :slightly_smiling_face:

**Additional context**
```
$ flutter test --coverage
mm:ss +N: All tests passed!

$ lcov --list coverage/lcov.info 
Reading tracefile coverage/lcov.info
                          |Lines       |Functions  |Branches    
Filename                  |Rate     Num|Rate    Num|Rate     Num
================================================================
[lib/src/]
[...]
foo.freezed.dart          |17.8%    416|    -     0|    -      0
[...]
================================================================
[...]
```

RounakTadvi:
This resolves #442 and #333.

i know this should be the role of source_gen instead, not Freezed, as this isn't Freezed-specific. But i guess until source_gen provides the solution to the issue, this would a great solution.

### DIFF
--- a/packages/freezed/lib/builder.dart
+++ b/packages/freezed/lib/builder.dart
@@ -7,6 +7,7 @@ import 'src/freezed_generator.dart';
 Builder freezed(BuilderOptions options) {
   return PartBuilder([FreezedGenerator(options.config)], '.freezed.dart',
       header: '''
+// coverage:ignore-file
 // GENERATED CODE - DO NOT MODIFY BY HAND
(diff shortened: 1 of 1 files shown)


rrousselGit:
Thanks

RounakTadvi:
You're welcome

PR #491, opened 2021-08-07
Support nullable unionKey

### LINKED ISSUE #467: Support null "unionKey" when a fallback is provided
How can we instruct freezed to type `json['type'] as String` to `json['type'] as String?` for this freezed configuration?

The left hand side file is `interaction.freezed.dart` while right-hand side is `interaction.dart` (freezed).

On line 38 on the right hand side, you can see that SectionType can be null. But I have to manually change `json['type'] as String` to `json['type']` or `json['type'] as String?` each time I rerun build-runner.

image

Thank you in advance!

Elias8:
Closes #467 by introducing a `nullableUnionKey` property to the `Freezed` annotation.

### DIFF
--- a/packages/freezed/lib/src/templates/from_json_template.dart
+++ b/packages/freezed/lib/src/templates/from_json_template.dart
@@ -49,7 +49,7 @@ class FromJson {
       }
 
       content = '''
-        switch (json['$unionKey'] as String) {
+        switch (json['$unionKey'] as String?) {
           $cases
           default:
             $defaultCase
--- a/packages/freezed/test/json_test.dart
+++ b/packages/freezed/test/json_test.dart
@@ -123,6 +123,19 @@ Future<void> main() async {
   });
 
   group('FreezedUnionFallback', () {
+    test(
+        'when fallback is non-null, if fromJson receives null as value, fallbacks to default',
+        () {
+      expect(
+        UnionFallback.fromJson(<String, dynamic>{'type': null, 'a': 42}),
+        UnionFallback.fallback(42),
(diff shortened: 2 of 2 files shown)


packages/freezed/lib/src/templates/from_json_template.dart
```diff
@@ -48,8 +50,9 @@ class FromJson {
             'return ${fallbackConstructor.redirectedName}$genericParameters.fromJson(json);';
       }
 
+      final nullabilitySuffix = nullableUnionKey ? '?' : '';
       content = '''
-        switch (json['$unionKey'] as String) {
+        switch (json['$unionKey'] as String$nullabilitySuffix) {
```
rrousselGit, line 55: Is there a use-case for having a fallbackUnion but have nullableUnionKey: false?

We could instead _always_ cast as `String?`

rrousselGit:
Thanks!

packages/freezed/lib/src/templates/from_json_template.dart
Elias8: Hmm, there is not.

We can always cast as `String?` when a `fallbackUnion` is not null.

Elias8: I thought the default case wasn't necessary since there is null safety. And I figured it can always be cast as `String?` since there is an exception thrown when a fallbackUnion is null. 🙂

Elias8:
Updated.

rrousselGit:
Epic thanks

Issue #451, opened 2021-05-22
Support @JsonSerializable(createFactory:false) 


moseskarunia:
**Problem**

- @JsonSerializable(createFactory:false) still generates _$classNameFromJson
- If I haven't write `factory ClassName.fromJson(Map<String, dynamic> json) =>_$ClassNameFromJson(json);` before running build_runner, it says that it cannot create toJson, and no .g.dart file is generated
- This is a nested object

**To Reproduce**

```dart
import 'package:freezed_annotation/freezed_annotation.dart';

part 'organization_param.freezed.dart';
part 'organization_param.g.dart';

@freezed
class PaginationOption with _$PaginationOption {
  @JsonSerializable(
    includeIfNull: false,
    checked: true,
    createFactory: false,
    explicitToJson: true,
  )
  const factory PaginationOption({
    required int queryLimit,
    String? lastObjectId,
  }) = _PaginationOption;

  factory PaginationOption.fromJson(Map<String, dynamic> json) =>
      _$PaginationOptionFromJson(json);
}

@freezed
class OrganizationQueryParam with _$OrganizationQueryParam {
  @JsonSerializable(
    includeIfNull: false,
    checked: true,
    explicitToJson: true,
    createFactory: false,
  )
  const factory OrganizationQueryParam({
    PaginationOption? paginationOpt,
  }) = _OrganizationQueryParam;

  factory OrganizationQueryParam.fromJson(Map<String, dynamic> json) =>
      _$OrganizationQueryParamFromJson(json);
}
```

To reproduce, try to remove either or both of fromJson factory code in those classes and run build_runner again. And there'll be this error:

```
[SEVERE] json_serializable:json_serializable on lib/params/organization_param.dart:

Could not generate `toJson` code for `paginationOpt`.
To support the type `PaginationOption` you can:
* Use `JsonConverter`
  https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonConverter-class.html
* Use `JsonKey` fields `fromJson` and `toJson`
  https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/fromJson.html
  https://pub.dev/documentation/json_annotation/latest/json_annotation/JsonKey/toJson.html
package:backoffice/params/organization_param.freezed.dart:286:27
    ╷
286 │   final PaginationOption? paginationOpt;
    │                           ^^^^^^^^^^^^^
    ╵
```

**Expected behavior**
[...]

PR #511, opened 2021-09-02
Support @JsonSerializable(createFactory:false) (#451)

koyama1003:
this addresses #451 
I've challenged myself to create a fix. Could you please have a moment to take a look at it?

# Changes
- Freezed won't create "fromJson" method if there is even one constructor annotated "@JsonSeriarizable(createFactory:false)"  exists.

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -79,7 +79,7 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
           GenericsDefinitionTemplate.fromGenericElement(element.typeParameters),
       genericsParameterTemplate:
           GenericsParameterTemplate.fromGenericElement(element.typeParameters),
-    );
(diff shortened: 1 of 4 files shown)


koyama1003:
My concern about my implementation is :

It won't generate "fromJson" method in the case of one constructor annotated "@JsonSeriarizable(createFactory:false)" exists, even if user wrote 'Class.fromJson' factory constructor.
Thus, an error is occured when  @JsonSeriarizable(createFactory:false) coexits with following  codes:

`factory Class.fromJson(Map<String, Object> json) => _$ClassFromJson(json);`

Is it better to dump a warning onto the terminal like this?
"We detected the createFactory:false attribute. So we won't generate fromJson method but only toJson method. To resolve this, nor delete the createFactory attribute or fromJson constructor."

rrousselGit:
Thanks for the help, but I'll close this for now. As mentioned in #451 I would prefer a Freezed-specific annotation instead of trying to decode `@JsonSerializable`, because that could easily become very complex.

Issue #517, opened 2021-09-14
Type signature issues with fromJson factory method tear-off


samandmoore:
**Describe the bug**
The `fromJson` tear-off included in the generated tear-off instance has the type signature `MyType fromJson(Map<String, Object> json)`. This is an issue because typically the `fromJson` factory's signature is `fromJson(Map<String, dynamic>)` or `fromJson(Map<String, Object?>)`. So, this tear-off is more restrictive than the actual factory constructor and it restricts the values that can be passed to the tear-off making it more or less unusable in places where you would want to use the factory constructor.

**To Reproduce**

Here's a class that shows the issue.

```dart
import 'package:freezed_annotation/freezed_annotation.dart';

part 'user.freezed.dart';
part 'user.g.dart';

@freezed
class User with _$User {
  factory User({
    required String name,
    int? age,
  }) = _User;

  factory User.fromJson(Map<String, dynamic> json) => _$UserFromJson(json);
}
```

And here's some code to consume it:

```dart
final Map<String, dynamic> someJson = <String, dynamic>{  'name': 'sam', age: 21 };
final user = $User.fromJson(someJson);
```

**Expected behavior**
I think we should change the type signature of the tear-off to `fromJson(Map<String, Object?>)` or `fromJson(Map<String, dynamic>)` depending on your preference.

Thanks for this library! It's awesome. I'm happy to open a PR with this change if you are onboard.

rrousselGit:
Hello!

That looks like a bug, indeed. Happy to review a PR :)
Chaing the Object to Object? should be fine

Le mar. 14 sept. 2021 à 14:34, Sam Moore ***@***.***> a
écrit :

samandmoore:
Awesome. Thanks for the guidance. 

Opened a PR up with a fix here: https://github.com/rrousselGit/freezed/pull/520

PR #520, opened 2021-09-19
fix(from-json-tear-off): adjust type signature of fromJson tear-off

samandmoore:
fixes https://github.com/rrousselGit/freezed/issues/517

this change swaps the type signature of the generated fromJson tear-off
to Map<String, Object?> instead of Map<String, Object>.

it seems most likely that this was missed in the upgrade to null-safety.
this essentially restores the previously intended behavior. maybe once
the next version of dart lands we won't need to generate these anymore
:) but for now it's super useful when it works!

**Note on testing**
this change doesn't update the tests because the file that tests this
is set to run with dart 2.9. i _think_ it's probably safe to change that test 
file to allow dart 2.12+, but i didn't wanna any unnecessary changes. 

fwiw if those tests were running with 2.12+ they'd have shown this issue:
image

### DIFF
--- a/packages/freezed/lib/src/templates/tear_off.dart
+++ b/packages/freezed/lib/src/templates/tear_off.dart
@@ -74,7 +74,7 @@ ${targetConstructor.redirectedName}$genericsParameter $ctorName$genericsDefiniti
 
     if (serializable) {
       yield '''
-$name$genericsParameter fromJson$genericsDefinition(Map<String, Object> json) {
+$name$genericsParameter fromJson$genericsDefinition(Map<String, Object?> json) {
   return $name$genericsParameter.fromJson(json);
 }
 ''';


samandmoore:
hey @rrousselGit if you have any time this week to take a look at this, it would be much appreciated! ❤️ 

thanks again for this package and all the others that make flutter so much more fun to work with.

rrousselGit:
Hello!
I'll look into it later this week :smile:

Feel free to ping me again if you see that I forgot

samandmoore:
hey again @rrousselGit 👋🏻 it looks like we both forgot about this last week. but consider this a friendly bump to take a look when you have a chance 😄

rrousselGit:
Sounds like it! I'll have a look tomorrow, while looking into another Freezed bug.

samandmoore:
howdy howdy howdy! i'm back with another polite bump on this @rrousselGit ❤️

PR #526, opened 2021-10-03
add support for custom == operator

casvanluijtelaar:
As discussed in https://github.com/rrousselGit/freezed/issues/273. I understand why you wouldn't want to promote it, but I feel that there is a good use for it in some scenarios and people who know what they're doing should be able to. 

I might need some guidance if it requires custom tests, I don't really understand the current structure.


casvanluijtelaar:
@rrousselGit I'm running into an issue when testing a custom equality operator in an implement class. I could probably fix it but I don't quite know how to debug a code gen package :) how do I use breakpoints / prints ?

casvanluijtelaar:
to get this straight, when there's a freezed class with a a `@With` decorator you expect that mixin's ==/hasCode to take priority over the class' own ==/hashCode? because that seems counterintuitive to me:

``` dart
mixin Foo {
  @override
  bool operator ==(Object o) => false;
}

class Bar with Foo {
  Bar(this.a);

  final String a;

  @override
  bool operator ==(Object o) => true;
}

void main() {
  print(Bar('ddd') == Bar('dfhf')); // true
}
```

rrousselGit:
Your example doesn't match what a Freezed class described.

The following:

```dart
mixin A {}

@freezed
class Union with _$Union {
  @With<A>()
  factory Union.first(int a) = UnionFirst;

  factory Union.second() = UnionSecond;
}
```

is equal to:

```dart
abstract class Union {}

mixin A on Union {}

class UnionFirst implements Union with A {}

class UnionSecond implements Union {}
```

In which case the mixin's `==` does take precedence over `Union`'s ==. 

What Freezed generates is `UnionFirst`/`UnionSecond`'s ==

rrousselGit:
I added an extends test and didn't have any issue. Maybe you had two classes/unions with the same name

rrousselGit:
In any case, LGTM

I'll merge when tests pass

casvanluijtelaar:
Thank you ;)

rrousselGit:
You did the job, I simply reviewed it :)

mariopepe:
hi @casvanluijtelaar I wanted to use this feature but I didn't quite understand how to do that, also looking at the MR code.

Would you please be so kind to share an example of how to define a custom == operator?

I will then study the example you'll provide, be sure to understand it, and I will contribute with a MR on the main readme so that other users in the future will be able to quickly find and use this great new feature of yours!

Thanks so much!

casvanluijtelaar:
> hi @casvanluijtelaar I wanted to use this feature but I didn't quite understand how to do that, also looking at the MR code.
> 
> Would you please be so kind to share an example of how to define a custom == operator?
> 
> I will then study the example you'll provide, be sure to understand it, and I will contribute with a MR on the main readme so that other users in the future will be able to quickly find and use this great new feature of yours!
> 
> Thanks so much!

Take a look at the examples folder 😉

mariopepe:
@casvanluijtelaar amazing!!! Sorry I don't know how I missed it!!! Thanks

PR #542, opened 2021-10-29
Properly handle class names with dollar sign `$`

comigor:
Freezed currently doesn't support the generation of classes with dollar sign character `$` (eg. `My$Class`) unless `toString` method is manually overridden because, due to the lack of escaping, code generation would consider `$Class` as a interpolation variable.

This quick fix simply escapes the character on both `toString` and `debugFillProperties` methods, where it's used inside a string.

### DIFF
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -206,7 +206,7 @@ Map<String, dynamic> toJson() {
 void debugFillProperties(DiagnosticPropertiesBuilder properties) {
   super.debugFillProperties(properties);
   properties
-    ..add(DiagnosticsProperty('type', '${constructor.fullName}'))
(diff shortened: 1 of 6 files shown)


packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -323,7 +323,7 @@ ${whenOrNullPrototype(allConstructors)} {
     return '''
 @override
 String toString($parameters) {
-  return '${constructor.fullName}(${properties.join(', ')})';
```
rrousselGit, line 326: Rather than dealing with escaping the name, a simpler solution would be to use:

```dart
return r'${constructor.fullName}...';
```

This way this will generate a string that doesn't interact with $

rrousselGit:
Thanks for this!

I'll quickly update to use `r'...'` instead of manual escapes and merge this

rrousselGit:
Actually you're right, it's simpler how you did it because the concatenation logic would otherwise be a bit tedious.

LGTM

comigor:
Thank you!

Issue #535, opened 2021-10-11
freezed 1.5.1+1 Code generation error with @Implements<Interface<Type>> annotation


lvsecoto:
**freezed 1.5.1+1 Code generation error with @Implements annotation**

If interface has `<Type>`，the generated code is incorrect.

Here is an example.

```
 abstract class SearchResultWord {
   String get title;
 
   String get definition;
 
   String get phonetic;
 }
 
 abstract class SearchResultDictWord<T extends TranslationResult>
     extends SearchResultWord {
   T get translation;
 }
 
 @freezed
 class SearchResultItem with _$SearchResultItem {
 
   @Implements<SearchResultDictWord<TranslationResultOffline>>()
   factory SearchResultItem.offlineDictWord({
     required String title,
     required String definition, 
     required String phonetic, 
     required TranslationResultOffline translation, 
   }) = SearchResultOfflineDictWord; 
}
```
Generated code:
```
abstract class SearchResultOfflineDictWord
    implements SearchResultItem, SearchResultDictWord<dynamic> {
  factory SearchResultOfflineDictWord(
          {required String title,
          required String definition,
          required String phonetic,
          required TranslationResultOffline translation}) =
      _$SearchResultOfflineDictWord;

  String get title;
  String get definition;
  String get phonetic;
  TranslationResultOffline get translation;
  @JsonKey(ignore: true)
  $SearchResultOfflineDictWordCopyWith<SearchResultOfflineDictWord>
      get copyWith => throw _privateConstructorUsedError;
}

```

`implements SearchResultItem, SearchResultDictWord<dynamic>` should be 
`implements SearchResultItem, SearchResultDictWord<TranslationResultOffline>`

rrousselGit:
Looking back at this, the problem stems from the fact that you are doing `@With<UnionCase>()`

In general I'd discourage against using the redirected class (`factory MyClass(..) = <<this>>`) as input of another class.
The analyzer will be unable to know that the generated class exist, so the code-generator will receive `dynamic`. That makes everything hellish to handle

rrousselGit:
Closing this as supporting this would be too difficult, and metaprogramming should fix such problem naturally.

PR #556, opened 2021-11-15
Use regex to parse With and Implements annotations

sroddy:
I propose a different way for parsing and outputting the `@With` and `@Implements` annotations that use the new "generic" approach.
The rationale is that we should be able to assume that the soundness (syntax error and well-formed types within the generic quotes) is guaranteed by the compiler itself, and whichever type is extracted by the regex can be copy-pasted, as it is, inside the generated file.

This approach would also solve edge-case situations like the one that surfaced in #535

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -299,16 +299,27 @@ Read here: https://github.com/rrousselGit/freezed/tree/master/packages/freezed#t
     return result;
   }
 
+  String _extractTypeFromGenericAnnotation(
+    String annotation,
+    ConstructorElement element,
+  ) {
+    final reg = RegExp(r'.*?<(.+)>');
+    final match = reg.firstMatch(annotation)?.group(1);
+    if (match == null) {
+      throw InvalidGenerationSourceError(
+        'Annotation does not properly specify type: $annotation',
+        element: element,
+      );
+    }
+
+    return match;
+  }
+
   Iterable<String> _withDecorationTypes(ConstructorElement constructor) sync* {
     for (final metadata in constructor.metadata) {
       if (!metadata.isWith) continue;
-      final object = metadata.computeConstantValue()!;
 
-      yield resolveFullTypeStringFrom(
-        constructor.library,
-        (object.type! as InterfaceType).typeArguments.single,
-        withNullability: false,
-      );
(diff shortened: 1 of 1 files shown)


rrousselGit:
Using `toSource` instead reliable / anti-pattern. For example folks could do:

```dart
const alias = With<Some<Complex<Type>>>();

@freezed
  @alias // that should work, but toSource would break this
  factory Example.first() = _ExampleFirst;
}
```

It should be possible to fix the bug mentioned without relying on manually parsing the source.

sroddy:
I think that the limitation that we have now (i.e. not being able to use any generated class inside the @With annotation) is much more painful than not supporting the complex use-case you posted above.

Of course I would love not to rely on a non-reliable implementation, but I'm a bit lost on how I could do that.
Do you have any idea/suggestion about how to do it?

rrousselGit:
The previous code relied on DartObject to obtain the generic type.
Chances are there are some extra informations I didn't use that contains the generics of the generic type used by the annotation

sroddy:
I've tried to play a bit with that when I was implementing this alternative approach and I failed finding an alternative like the one you are suggesting. The problem is that if the generic is tried to be resolved in any way by the analyzer package, and the type is generated by freezed, it gets resolved as `dynamic`. Of course I may be missing something, but it's very hard for me to completely understand how the `analyzer` package really works as its somewhat hard to find docs and examples that uses it in a complex scenario like the freezed one

rrousselGit:
>  The problem is that if the generic is tried to be resolved in any way by the analyzer package, and the type is generated by freezed, it gets resolved as dynamic.

I'm alright with that.

To begin with you shouldn't use a Freezed class inside a `@With`/`@Implements`.

sroddy:
>To begin with you shouldn't use a Freezed class inside a @With/@Implements.

Why? Unfortunately Dart doesn't let you define an interface that contains existential types (for example that contains a method which returns Self/This), so it's pretty common to have interfaces that are generic to the specific instance that implements the interface itself in order to overcome that issue.

https://github.com/dart-lang/sdk/issues/28477

BTW, this was a perfectly supported scenario until v0.15 and I think it's quite a pity that it can no longer be supported.

rrousselGit:
I don't like using regex here. What do you think about adding the `fromString` constructor back? We could have `@WithFromString('MyClass')`

rrousselGit:
:wave:

I don't think I'll go forward with this change. A regex for parsing the source code is IMO dangerous

PR #583, opened 2022-01-06
Add option `fromJsonFactory` and `toJson` to toogle code generations

morlay:
fix https://github.com/rrousselGit/freezed/issues/451

### DIFF
--- a/packages/_internal/lib/models.dart
+++ b/packages/_internal/lib/models.dart
@@ -65,7 +65,7 @@ class ConstructorDetails with _$ConstructorDetails {
 }
 
 @freezed
-class Data with _$Data {
+abstract class Data with _$Data {
   @Assert('constructors.isNotEmpty')
   factory Data({
     required String name,
@@ -80,6 +80,8 @@ class Data with _$Data {
     required bool hasCustomEquals,
     required bool shouldGenerateMaybeMap,
     required bool shouldGenerateMaybeWhen,
+    required bool shouldGenerateFromJsonFactory,
+    required bool shouldGenerateToJson,
   }) = _Data;
 }
 
--- a/packages/_internal/lib/models.freezed.dart
+++ b/packages/_internal/lib/models.freezed.dart
@@ -752,7 +752,9 @@ class _$DataTearOff {
       required bool hasCustomToString,
       required bool hasCustomEquals,
       required bool shouldGenerateMaybeMap,
-      required bool shouldGenerateMaybeWhen}) {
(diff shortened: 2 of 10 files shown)


rrousselGit:
I do not want to make Freezed dependent on how json_serializable works.

If we're going to disable the fromJson generation, this should be parameterized using Freezed options instead of json_serializable options

morlay:
@rrousselGit Yes. 

But in fact, Freezed depend `fromJson` which generated by `json_serializable`;
When `@JsonSerializable(createFactory: false)`, Freezed will generate wrong codes with unknown "xxxFromJSON" function.

For `toJson`, could be disabled by another way. 
When custom `toJson` defined,  don't generate it.

rrousselGit:
Right, because disabling the generation of the factory is not supported.

If we want to support this, it shouldn't be through `@JsonSerializable(createFactory: false)` but a Freezed annotation instead.

morlay:
@rrousselGit

But in that way, it will change a lot.
need consider merge `createFromJson` and `createToJson` of `Freezed` into `JsonSerializable`.
Configurate twice is terrible.

morlay:
@rrousselGit  if we decide to add options of `Freezed`.

How about `@freezed(createFromJsonFactory: false, createToJson: false)` ?

I will try to refact this PR to got it.

rrousselGit:
By default, users do not specify the `@JsonSerializable` annotation.

So when we do `@Freezed(fromJson: false)`, then the generator can generate `@JsonSerializable(createFactory: false)`

morlay:
@rrousselGit updated. 

Follow the option styles, now it will be:
`@Freezed(fromJsonFactory: false, toJson: false)`
it could disable only one of  `fromJsonFactory` or `toJson` like `@JsonSerializable` did.

morlay:
Hi @rrousselGit
is there other concerns?

yh-luo:
This implementation is really useful for creating data models from API responses. Currently, I need to either manually remove all `toJson` in `*.freezed.dart`, or give up on `create_to_json: false`.
Would this PR be merged?

Or maybe `freezed` is not suitable for that use case? (but I really like the ==/toString implementations)

sadi304:
This is getting merged? useful for the coverage as some apis are only get only

rrousselGit:
Closing as I'll take care of this in a different PR

I'll revamp this to allow disabling _anything_ instead of only fromJson/toJson

Your work is very appreciated though :smile:

PR #590, opened 2022-01-24

TimWhiting:
@fabianMendez, @rrousselGit 

I hope you don't mind but I took #370, and resolved the outdated bits. I believe it is ready to merge & I definitely could use it in a few of my projects.

Was there anything that was preventing the merge before? It seems like it was approved before, but got out of sync with master.


serendipity1004:
O I want this merged so badly

TimWhiting:
Unfortunately, this will required some rework, since the master branch has diverged quite a bit.

TimWhiting:
I ended up starting from scratch: See #696

PR #602, opened 2022-02-20
Fix "information" plural typo

nilsreichardt:
In English, there is no plural for "information". It's just information.

> We do not use information in the plural form and we do not use it with a/an. We use piece or pieces to make information countable:
I found a lot of information about Ecuador on the Internet.
Not: I found a lot of informations about Ecuador …

https://dictionary.cambridge.org/grammar/british-grammar/information


rrousselGit:
Good to know!

PR #603, opened 2022-02-20
doc: fix document typo

thinhlh:
I would recommend to fix this at other branches.


rrousselGit:
This isn't a mistake. "impossible" is the correct word here

PR #643, opened 2022-04-18
Optimize collection ==/hashCode for when they do not need a "DeepCollectionEquality"

### LINKED ISSUE #626: `operator==` is terribly slow for lists - can we customize it? or change its implementation?
Hi thanks for the lib! However, I am using it for large arrays(`List<int>`, or more specifically, `Uint8List`), and the following:

```
  const factory SomeClass.bytes(List<int> bytes) = SomeClassModeBytes;
```

generates:

```
  @override
  bool operator ==(dynamic other) {
    return identical(this, other) ||
        (other.runtimeType == runtimeType &&
            other is SomeClassModeBytes &&
            const DeepCollectionEquality().equals(other.bytes, bytes));
  }
```

After digging into `DeepCollectionEquality`, it delegates to `ListEquality`, and the core code for the latter is:

```
    for (var i = 0; i < length; i++) {
      if (!_elementEquality.equals(list1[i], list2[i])) return false;
    }
```

As we all know, this is *terribly slow*. There are tons of method calls there, and the types are `dynamic` so it cannot know it is a `int` in advance so has to do tons of work instead of a simple `int` equality check.

Can we customize the generated `operator==`? Or, can `freezed` be improved, such that it uses `listEquals` https://api.flutter.dev/flutter/foundation/listEquals.html instead? (That method is indeed about 10 lines of code, so it is even easy to copy-and-paste into freezed's library)

fzyzcjy:
Close #626

### DIFF
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -374,7 +374,6 @@ String toString($parameters) {
 ''';
   }
 
-  // TODO
   String get _operatorEqualMethod {
     if (!data.generateEqual) return '';
 
@@ -384,8 +383,8 @@ String toString($parameters) {
       ...constructor.impliedProperties.map((p) {
         final name = p.name == 'other' ? 'this.other' : p.name;
         if (p.isPossiblyDartCollection) {
-          // no need to check `identical` as `DeepCollectionEquality` already does it
-          return 'const DeepCollectionEquality().equals(other.${p.name}, $name)';
+          // no need to check `identical` as `FastDeepCollectionEquality` already does it
+          return 'const FastDeepCollectionEquality().equals(other.${p.name}, $name)';
         }
         return '(identical(other.${p.name}, $name) || other.${p.name} == $name)';
       }),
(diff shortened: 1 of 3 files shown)


rrousselGit:
Hum interesting. That's not quite what I would've done

Since we're using code-generation, I was thinking of using the analyzer to detect "Hey this is a List<num|Enum|String|bool>, let's do the == using listEquals instead"

This would not have any impact on List<Object>

Which means that instead of changing `DeepCollectionEquality`, we'd change the generated code instead.

rrousselGit:
In that sense, this is an optimization that is only possible because we're a code-generator.

Also, if in the end, your version of DeepCollectionEquality is indeed faster, then I think this change should not be a PR to Freezed but instead a PR to package:collection.

fzyzcjy:
Hmm I know that is another possible approach that comes to my mind. But IMHO that is not very trivial to implement and needs quite a bit of type checks. In addition, since we are using `DeepCollectionEquality`, we already have some performance deduction because its implementation has things like `if (a is Set)` and so on, so our extra two `if (a is List<int>)` should not be expensive.

> I think this change should not be a PR to Freezed but instead a PR to package:collection.

I will make a PR there as well.

rrousselGit:
>  But IMHO that is not very trivial to implement and needs quite a bit of type checks

It shouldn't be too difficult actually

It's a matter of `dartType.isDartCoreList && (dartType as InterfaceType).typeArguments.single == <check is int>`

fzyzcjy:
I see. I will do that when having time.

rrousselGit:
Closing. This change shouldn't be done on Freezed, but package:collection instead

If Freezed were to do such performance optimization, it should instead be about not using `DeepCollectionEquality` to begin with

PR #696, opened 2022-06-28

TimWhiting:
New attempt at resolving https://github.com/rrousselGit/freezed/pull/370
fixes: #331


rrousselGit:
Sorry for the delay. I'll look into it

packages/freezed/test/integration/json.dart
```diff
@@ -452,7 +452,7 @@ class Generic<T> with _$Generic<T> {
   factory Generic(@DataConverter() T a) = _Generic<T>;
 
   factory Generic.fromJson(Map<String, dynamic> json) =>
-      _$GenericFromJson<T>(json);
+      _$GenericFromJson<T>(json, (DataConverter<T>()).fromJson);
```
TimWhiting, line 455: This currently is a negative change in my opinion.
Two ways to resolve this.
1. Automatically detect JsonConverter annotations that handle generic types matching the surrounding class. (Very difficult)
2. Explicit opt in / out of generic argument json factories, much easier, but unfortunate both if we make it opt in or opt out, and I'm not sure which I would prefer.

TimWhiting:
> Sorry for the delay. I'll look into it

Thanks, this is something I've wanted several times. This is a much cleaner approach than before, and rather simple, probably because of the rearchitecture of the freezed code you did recently. I've highlighted one issue that I see with my current solution. I'd like your opinion on how to proceed there. I've attempted the first solution separate from this PR, but I found it rather difficult to correlate type parameters on specific annotations of constructor arguments to the generic parameter of the surrounding class.

packages/freezed/test/integration/json.dart
rrousselGit: What is the difficulty with 1 in your opinion?

packages/_internal/lib/models.freezed.dart
```diff
@@ -11,7 +11,7 @@ part of 'models.dart';
 T _$identity<T>(T value) => value;
 
 final _privateConstructorUsedError = UnsupportedError(
-    'It seems like you constructed your class using `MyClass._()`. This constructor is only meant to be used by freezed and you are not supposed to need it nor use it.\nPlease check the documentation here for more information: https://github.com/rrousselGit/freezed#custom-getters-and-methods');
+    'It seems like you constructed your class using `MyClass._()`. This constructor is only meant to be used by freezed and you are not supposed to need it nor use it.\nPlease check the documentation here for more informations: https://github.com/rrousselGit/freezed#custom-getters-and-methods');
```
rrousselGit, line 14:
```suggestion
    'It seems like you constructed your class using `MyClass._()`. This constructor is only meant to be used by freezed and you are not supposed to need it nor use it.\nPlease check the documentation here for more information: https://github.com/rrousselGit/freezed#custom-getters-and-methods');
```

packages/freezed/test/integration/json.dart
rrousselGit: The fact that this is technically a breaking change bothers me. 
I think we'd need to be opt-in.

rrousselGit:
We'd need a matching parameter on the annotation and a config in the build.yaml file

TimWhiting:
Updated to the opt-in solution with a parameter in the annotation and config in the build.yaml file.

packages/freezed/test/integration/json.dart
TimWhiting: The difficulty is that you have to make sure that **every** parameter in **every** constructor has a type converter for any type recursively containing **any** generic type of the class, even for nested types such as a parameter that has type `List<T>` would need a type converter for `List<T>`.

rrousselGit:
Could you add doc for this? Including the build.yaml config, which likely will be used often

TimWhiting:
@rrousselGit Added to the readme

TimWhiting:
Any more feedback, or can this get merged?

rrousselGit:
Nothing to say, thanks for your work!

PR #705, opened 2022-07-10
[Update] Korean translation

byunggilchoi:
- [Fix] correct some mistranslations about union type
- [Fix] rephrase some korean sentences


rrousselGit:
Thanks!

PR #735, opened 2022-08-14
Improve the performance of copyWith

miDeb:
This PR aims to improve the performance of `copyWith` calls, even though performance parity with a naive implementation was still not achieved.

To measure performance I added a benchmark to the benchmarks folder. Below you can see how each change improved performance:

|Change|Runtime|Improvement|
|--|--|--|
|**No change**|**595 ms**| - |
|Inlining|479 ms|19%|
|Comparison order|420 ms|12%|
|`null` instead of `freezed` as the default value|405 ms|3.5%|
|avoiding downcasts|175 ms|56%|
|**Total**|**175ms**|**70%**|

As you can see, for this particular benchmark the runtime was decreased by about 70%. Benchmarks are obviously not representative of the real world, but we can at least assume that there will be **some** improvement.

However, even with these changes the compiler still doesn't completely understand what we're trying to do. A naive implementation of this benchmark will complete immediately because the compiler understands that we're essentially just incrementing a counter. More investigation would be needed to figure out how we can make the compiler treat a `copyWith` call like a simple constructor invocation. Nonetheless I think these changes already improve performance substantially.


rrousselGit:
Some of these changes are breaking changes.

We'd need tests to pass for that to be reasonable

miDeb:
Sorry, I missed those failing tests, I'll try to fix them.

rrousselGit:
In particular, using `null` is a big no-no. Which prevents using a downcast too

Freezed explicitly supports `copyWith(property: null)`

rrousselGit:
Wait, I may have spoken too early as I didn't look at your code

If you did indeed change only non-nullable properties, that should be fine :)

miDeb:
You were right that some tests weren't passing, which I had missed. But I don't think there is any inherently breaking change in this PR.

PR #736, opened 2022-08-14
Add `useResult` to copyWith calls

### LINKED ISSUE #734: Add `@useResult` to copyWith
Sometimes when you get distracted you might forget to actually assign the result of calling `copyWith` to the variable you want to change. While an issue like this is obvious once you spot it, doing so may take some time. Also, beginners who are not yet familiar with the concept of `copyWith` may try to use it incorrectly.

[Starting from dart 2.18 ](https://github.com/dart-lang/sdk/issues/48913) the analyzer will warn you if you do not use the result of calling a callable class whose `call` method is annotated with `@useResult`. This seems like a good opportunity to start using this annotation in `freezed`.

I don't think there are any alternatives apart from just not making this change.

**Additional context**
If this is a feature you want I'm ready to implement it.

miDeb:
Closes #734.
[Starting from dart 2.18 ](https://github.com/dart-lang/sdk/issues/48913) the analyzer will warn you if you do not use the result of calling a callable class whose `call` method is annotated with `@useResult`.

This PR adds `@useResult` annotations to `copyWith` calls.


miDeb:
It seems like the flutter master channel formats some files differently than the stable channel. To make the check happy I formatted everything using the master channel formatter.

PR #740, opened 2022-08-15
feat: add common properties with different nullability to base mixin

### LINKED ISSUE #715: Allow reading common properties that share a parent class/ implementation
**The problem:**
We can currently read properties that are defined on all constructors as documented [here](https://github.com/rrousselGit/freezed#shared-properties).
However, when a property shares a name but not the type we lose the ability to read those properties. This is true even when the property shares a parent class. Pattern matching is required to read the property when the user may have only wanted the members of its parent class.

Consider:
```dart
abstract class SomeInterface{}

class A implements SomeInterface{}
class B implements SomeInterface{}

  const factory Example.someCase(A commonProperty) = SomeCase;
  const factory Example.anotherCase(B commonProperty) = AnotherCase;
  const factory Example.aThirdCase(SomeInterface commonProperty) = ThirdCase;
}

final Example example = ...
final SomeInterface commonProperty = example.commonProperty; //does not compile
```

The above is not possible despite each case containining a property that share a name ("commonProperty") and implementation (```abstract class SomeInterface```):

Instead, pattern matching is required to read commonProperty. There may be contexts where we don't care about the union case and merely want to read the shared property as SomeInterface.

**An example API suggestion**

```dart
  @CommonClass<SomeInterface>("commonProperty")  
  const factory Example.someCase(A commonProperty) = SomeCase;
  const factory Example.anotherCase(B commonProperty) = AnotherCase;
  const factory Example.aThirdCase(SomeInterface commonProperty) = ThirdCase;
}
```

Which would then allow
```dart
final Example example = ...
final SomeInterface commonProperty = example.commonProperty;
```

Would something like this be possible?

DevNico:
The new field `isCommonWithDifferentNullability` might not be perfect but it was the best way I could come up with to propagate the information.

Close #715

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -147,6 +147,8 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
           isPossiblyDartCollection: commonParameter.isPossiblyDartCollection,
           // TODO: support hasJsonKey
           hasJsonKey: false,
+          isCommonWithDifferentNullability:
+              commonParameter.isCommonWithDifferentNullability,
         ),
     ];
   }
@@ -252,25 +254,48 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
   ) {
     return constructorsNeedsGeneration.first.parameters.allParameters
         .map((parameter) {
-          var hasAnyFinalProperty = false;
+          var anyMatchingPropertyIsFinal = false;
+          var anyMatchingPropertyIsNullable = false;
+
           for (final constructor in constructorsNeedsGeneration) {
             final matchingParameter =
                 constructor.parameters.allParameters.firstWhereOrNull((p) {
-              return p.name == parameter.name && p.type == parameter.type;
+              return p.name == parameter.name &&
+                  _typeStringWithoutNullability(p.type) ==
+                      _typeStringWithoutNullability(parameter.type);
             });
 
             if (matchingParameter == null) return null;
-            if (matchingParameter.isFinal) hasAnyFinalProperty = true;
+            if (matchingParameter.isFinal) anyMatchingPropertyIsFinal = true;
+            if (matchingParameter.isNullable)
+              anyMatchingPropertyIsNullable = true;
           }
 
-          if (hasAnyFinalProperty) {
-            return parameter.copyWith(isFinal: true);
+          final isNullable =
+              parameter.isNullable || anyMatchingPropertyIsNullable;
+
+          if (parameter.isNullable != isNullable) {
+            print(parameter.name);
           }
 
-          return parameter;
+          return parameter.copyWith(
+            isFinal: parameter.isFinal || anyMatchingPropertyIsFinal,
+            isNullable: isNullable,
+            type: isNullable && (parameter.type?.endsWith('?') == false)
+                ? '${parameter.type}?'
+                : parameter.type,
+            isCommonWithDifferentNullability:
+                parameter.isNullable != isNullable,
+          );
         })
         .whereNotNull()
         .toList();
+  }
+
+  String? _typeStringWithoutNullability(String? type) {
+    return type?.endsWith('?') == true
+        ? type?.substring(0, type.length - 1)
+        : type;
   }
 
   Future<List<ConstructorDetails>> _parseConstructorsNeedsGeneration(
@@ -606,7 +631,9 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
             ).toList(),
             genericsDefinition: data.genericsDefinitionTemplate,
             genericsParameter: data.genericsParameterTemplate,
-            allProperties: commonProperties,
+            allProperties: commonProperties
+                .whereNot((element) => element.isCommonWithDifferentNullability)
+                .toList(),
             data: data,
           );
 
--- a/packages/freezed/lib/src/templates/copy_with.dart
+++ b/packages/freezed/lib/src/templates/copy_with.dart
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
 import 'package:freezed/src/templates/parameter_template.dart';
 import 'package:freezed/src/templates/properties.dart';
 
@@ -65,23 +66,28 @@ ${_abstractDeepCopyMethods().join()}
       final body = _copyWithMethodBody(
         parametersTemplate: ParametersTemplate(
           const [],
-          namedParameters: commonProperties.map((e) {
-            return Parameter(
-              decorators: e.decorators,
-              name: e.name,
-              isNullable: false,
-              isFinal: false,
-              isDartList: false,
-              isDartMap: false,
-              isDartSet: false,
-              showDefaultValue: false,
-              isRequired: false,
-              defaultValueSource: '',
-              type: e.type,
-              doc: e.doc,
-              isPossiblyDartCollection: e.isPossiblyDartCollection,
-            );
-          }).toList(),
+          namedParameters: commonProperties
+              .map((e) {
+                return Parameter(
+                  decorators: e.decorators,
+                  name: e.name,
+                  isNullable: false,
+                  isFinal: false,
+                  isDartList: false,
+                  isDartMap: false,
+                  isDartSet: false,
+                  showDefaultValue: false,
+                  isRequired: false,
+                  defaultValueSource: '',
+                  type: e.type,
+                  doc: e.doc,
+                  isPossiblyDartCollection: e.isPossiblyDartCollection,
+                  isCommonWithDifferentNullability:
+                      e.isCommonWithDifferentNullability,
+                );
+              })
+              .whereNot((element) => element.isCommonWithDifferentNullability)
+              .toList(),
         ),
         returnType: '_value.copyWith',
       );
--- a/packages/freezed/lib/src/templates/properties.dart
+++ b/packages/freezed/lib/src/templates/properties.dart
@@ -31,6 +31,7 @@ class Property {
     required this.isDartMap,
     required this.isDartSet,
     required this.isPossiblyDartCollection,
+    required this.isCommonWithDifferentNullability,
   }) : type = type ?? 'dynamic';
 
   static Future<Property> fromParameter(
@@ -60,6 +61,7 @@ class Property {
       defaultValueSource: defaultValue,
       hasJsonKey: element.hasJsonKey,
       isPossiblyDartCollection: element.type.isPossiblyDartCollection,
+      isCommonWithDifferentNullability: false,
     );
   }
 
@@ -75,6 +77,7 @@ class Property {
   final bool hasJsonKey;
   final String doc;
   final bool isPossiblyDartCollection;
+  final bool isCommonWithDifferentNullability;
 
   @override
   String toString() {
@@ -135,6 +138,7 @@ class Property {
     bool? hasJsonKey,
     String? doc,
     bool? isPossiblyDartCollection,
+    bool? isCommonWithDifferentNullability,
   }) {
     return Property(
       type: type ?? this.type,
@@ -150,6 +154,8 @@ class Property {
       isDartSet: isDartSet ?? this.isDartSet,
       isPossiblyDartCollection:
           isPossiblyDartCollection ?? this.isPossiblyDartCollection,
+      isCommonWithDifferentNullability: isCommonWithDifferentNullability ??
+          this.isCommonWithDifferentNullability,
     );
   }
 }
--- a/packages/freezed/test/multiple_constructors_test.dart
+++ b/packages/freezed/test/multiple_constructors_test.dart
@@ -504,8 +504,48 @@ void main() {
 }
 '''), throwsCompileError);
     });
+  });
 
-    // TODO: shared property name but different type
+  group('SharedParamNullable', () {
+    test('has the common properties available', () {
+      SharedParamNullable value = SharedParamNullable('a', 'b', 42);
+      expect(value.a, 'a');
+
+      value = SharedParamNullable.named('b', 'c', 24);
+      expect(value.a, 'b');
+    });
+
+    test(
+      "copy doesn't have shared params with different nullability",
+      () async {
+        await expectLater(compile(r'''
+import 'multiple_constructors.dart';
+
+void main() {
+  final param = SharedParamNullable('a', 'b', 42);
+  param.copyWith(a: '2');
+}
+'''), throwsCompileError);
+
+        await expectLater(compile(r'''
+import 'multiple_constructors.dart';
+
+void main() {
+  final param = SharedParamNullable('a', 'b', 42);
+  param.copyWith(b: '1');
+}
+'''), completes);
+
+        await expectLater(compile(r'''
+import 'multiple_constructors.dart';
+
+void main() {
+  final param = SharedParamNullable('a', 'b', 42);
+  param.copyWith(c: 42);
+}
+'''), throwsCompileError);
+      },
+    );
   });
 
   test('Can have a named constructor and a property using the same name', () {
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -135,6 +135,7 @@ ${copyWith?.abstractCopyWithGetter ?? ''}
         doc: '',
         isPossiblyDartCollection: false,
         showDefaultValue: false,
+        isCommonWithDifferentNullability: false,
       );
 
       parameters = ParametersTemplate(
--- a/packages/freezed/lib/src/templates/parameter_template.dart
+++ b/packages/freezed/lib/src/templates/parameter_template.dart
@@ -82,6 +82,7 @@ class ParametersTemplate {
         doc: await documentationOfParameter(e, buildStep),
         isPossiblyDartCollection: e.type.isPossiblyDartCollection,
         showDefaultValue: true,
+        isCommonWithDifferentNullability: false,
       );
 
       if (isAssignedToThis) return LocalParameter.fromParameter(value);
@@ -184,6 +185,7 @@ class Parameter {
     required this.isFinal,
     required this.isPossiblyDartCollection,
     required this.showDefaultValue,
+    required this.isCommonWithDifferentNullability,
   });
 
   Parameter.fromParameter(Parameter p)
@@ -201,6 +203,7 @@ class Parameter {
           showDefaultValue: p.showDefaultValue,
           doc: p.doc,
           isPossiblyDartCollection: p.isPossiblyDartCollection,
+          isCommonWithDifferentNullability: p.isCommonWithDifferentNullability,
         );
 
   final String? type;
@@ -216,6 +219,7 @@ class Parameter {
   final bool isPossiblyDartCollection;
   final bool isFinal;
   final String doc;
+  final bool isCommonWithDifferentNullability;
 
   Parameter copyWith({
     String? type,
@@ -232,6 +236,7 @@ class Parameter {
     bool? isDartMap,
     bool? isDartSet,
     bool? isFinal,
+    bool? isCommonWithDifferentNullability,
   }) =>
       Parameter(
         type: type ?? this.type,
@@ -248,6 +253,8 @@ class Parameter {
         isFinal: isFinal ?? this.isFinal,
         isPossiblyDartCollection:
             isPossiblyDartCollection ?? this.isPossiblyDartCollection,
+        isCommonWithDifferentNullability: isCommonWithDifferentNullability ??
+            this.isCommonWithDifferentNullability,
       );
 
   @override
@@ -283,6 +290,7 @@ class LocalParameter extends Parameter {
     required List<String> decorators,
     required String doc,
     required bool isPossiblyDartCollection,
+    required bool isCommonWithDifferentNullability,
   }) : super(
           name: name,
           type: type,
@@ -297,6 +305,7 @@ class LocalParameter extends Parameter {
           defaultValueSource: defaultValueSource,
           doc: doc,
           isPossiblyDartCollection: isPossiblyDartCollection,
+          isCommonWithDifferentNullability: isCommonWithDifferentNullability,
         );
 
   LocalParameter.fromParameter(Parameter p)
@@ -313,6 +322,7 @@ class LocalParameter extends Parameter {
           decorators: p.decorators,
           doc: p.doc,
           isPossiblyDartCollection: p.isPossiblyDartCollection,
+          isCommonWithDifferentNullability: p.isCommonWithDifferentNullability,
         );
 
   @override
@@ -346,6 +356,7 @@ class CallbackParameter extends Parameter {
     required this.parameters,
     required String doc,
     required bool isPossiblyDartCollection,
+    required bool isCommonWithDifferentNullability,
   }) : super(
           name: name,
           type: type,
@@ -360,6 +371,7 @@ class CallbackParameter extends Parameter {
           defaultValueSource: defaultValueSource,
           doc: doc,
           isPossiblyDartCollection: isPossiblyDartCollection,
+          isCommonWithDifferentNullability: isCommonWithDifferentNullability,
         );
 
   final ParametersTemplate parameters;
--- a/packages/freezed/lib/src/templates/prototypes.dart
+++ b/packages/freezed/lib/src/templates/prototypes.dart
@@ -127,6 +127,7 @@ String _mapPrototype(
           // TODO: do we want to support freezed classes that implements MapView/ListView?
           isPossiblyDartCollection: false,
           showDefaultValue: false,
+          isCommonWithDifferentNullability: false,
         ),
       ]);
     },
@@ -191,6 +192,7 @@ String _unionPrototype(
       defaultValueSource: '',
       doc: '',
       isPossiblyDartCollection: false,
+      isCommonWithDifferentNullability: false,
     );
 
     if (constructor.isDefault) {
--- a/packages/freezed/pubspec.yaml
+++ b/packages/freezed/pubspec.yaml
@@ -6,10 +6,10 @@ version: 2.1.0+1
 homepage: https://github.com/rrousselGit/freezed
 
 environment:
-  sdk: '>=2.17.0 <3.0.0'
+  sdk: ">=2.17.0 <3.0.0"
 
 dependencies:
-  analyzer: ">=4.2.0 <5.0.0"
+  analyzer: ">=4.6.0 <5.0.0"
   build: ^2.0.0
   build_config: ^1.0.0
   collection: ^1.15.0
--- a/packages/freezed/test/integration/multiple_constructors.dart
+++ b/packages/freezed/test/integration/multiple_constructors.dart
@@ -65,6 +65,14 @@ class SharedParam with _$SharedParam {
 }
 
 @freezed
+class SharedParamNullable with _$SharedParamNullable {
+  const factory SharedParamNullable(String a, String b, int c) =
+      SharedParamNullable0;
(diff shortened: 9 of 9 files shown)


DevNico:
I just also had the thought that we could also expose non-common properties that don't have a type collision as nullable on the base mixin. What do you think? @rrousselGit

DevNico:
In theory, we could also add all properties that are not common in all constructors and do not have a type conflict to the base mixin.

What do you think? @rrousselGit?

DevNico:
Also instead of removing properties with different nullability, we could add them but only allow non-null inputs.

DevNico:
> Also instead of removing properties with different nullability, we could add them but only allow non-null inputs.

I went ahead and implemented that feature as well

TimWhiting:
I don't think this closes #715 which is requesting for a common base class, whereas this just addresses the nullability aspect. This is a great first step though!

DevNico:
You're right I read that issue a while ago and needed this specific feature now so I went ahead and implemented it. I vaguely remembered reading something about it and therefore just quickly added the closing statement without re-reading the original issue.

Removed the closing statement since it is out of scope for what I currently require and would also require a lot more work.

rrousselGit:
Hello! Thanks for your work!


Supporting cases like:

```dart
factory Foo.first(int a) = _First;
factory Foo.second(double a) = _First;
...


Foo foo;
num a = foo.a;
```

should not be as complex as I originally thought.


I later learned about `TypeSystem`/`TypeProvider` (which can be found on `LibraryElement` I believe). They provide utilities find the nearest common type between two properties

It's probably worth using that directly instead.

DevNico:
If you haven't started working on that I can give it a go and try to refactor this PR to use `TypeSystem`/`TypeProvider`

rrousselGit:
Sure, go ahead 😄

DevNico:
Properties with a common super type that isn't `Object` now have getters in the base mixin. 🎉

I have not included them in the `copyWith` since we can only upcast to the common supertype but not downcast.

DevNico:
I also noticed that the `super` constructors could be refactored to make use of the `super.` shorthand. Is that something you'd be alright with? I could do that in this PR or open another one for those changes.

rrousselGit:
Great!

The super syntax isn't used yet because Freezed supports an sdk version lower than 2.17

It's not very important considering the code is generated

rrousselGit:
I'm a bit busy until the Flutter Viking talks the 1st of September. Not sure I'll be able to properly review this PR before

But thanks a lot for your work 😊

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -248,26 +240,48 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
...
 
             if (matchingParameter == null) return null;
-            if (matchingParameter.isFinal) hasAnyFinalProperty = true;
-          }
 
-          if (hasAnyFinalProperty) {
-            return parameter.copyWith(isFinal: true);
+            final commonTypeWithMatch =
+                parameter.parameterElement!.library!.typeSystem.leastUpperBound(
+              parameter.parameterElement!.type,
```
TimWhiting, line 261: Should this be commonType? (Initialize commonType with parameter.parameterElement!.type). That way it can handle >2 constructors with a common base type

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -248,26 +240,48 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
...
+              matchingParameter.parameterElement!.type,
+            );
+
+            if (commonTypeWithMatch.isDartCoreObject) return null;
```
TimWhiting, line 265: Why return early here? Object is more specific than Object? (it gives you a non-null object, but I guess there isn't much you can do with Object). I would assume that if someone named all the parameters the same there could be a case made for wanting access to that parameter even if it returns Object?

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -248,26 +240,48 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
...
+
+            if (!allowCommonSuperType &&
+                commonTypeWithMatch.getDisplayString(withNullability: false) !=
+                    parameter.type?.replaceAll(r'?$', '')) return null;
```
TimWhiting, line 269: Why not `parameter.parameterElement.type.getDisplayString(withNullability: false)`, instead of manually removing nullability? Additionally shouldn't the `!allowCommonSuperType` check occur for determining the `commonType` as well? Specifically in the copyWith below here you might want to only use `commonType` if `allowCommonSuperType` is true.

packages/freezed/lib/src/freezed_generator.dart
TimWhiting: You could then remove the if (commonType != commonTypeWithMatch) check below.

packages/freezed/test/integration/single_class_constructor.dart
```diff
@@ -450,7 +450,7 @@ class NullDefault with _$NullDefault {
 class ExplicitConstDefault with _$ExplicitConstDefault {
   factory ExplicitConstDefault(
       //ignore: unnecessary_const
-      [@Default(const <Object>[]) List<Object> value]) = _ExplicitConstDefault;
```
TimWhiting, line 453: Why was this removed? As you can see with the lint, and the name of this class, this is testing support for the explicit const keyword here.

packages/freezed/test/multiple_constructors_test.dart
```diff
@@ -504,8 +504,65 @@ void main() {
...
+    test(
+      'copy does not have shared params with common super type available',
+      () async {
+        await expectLater(compile(r'''
+import 'multiple_constructors.dart';
+
+void main() {
+  final param = SharedParamCommonSuperType('a', 'b', 42);
+  param.copyWith(b: 42);
+}
```
TimWhiting, line 562: This probably could be done, (not a common super type, but by finding a common sub type).

For example in the null-case:
String?
String
common subtype == String, which is why copyWith a String makes sense because it works for all variations of the parameter.

For a not much more complicated hierarchy.
num
double
common subtype == double -> you can use double for the parameter

Finally, in general:
if (param is subtype of otherConstructorParam):
  commonSubtype = param
else if (otherConstructorParam is subtype of param):
  commonSubtype = otherConstructorParam
else
  commonSubtype = Never (no common subtype)


Additionally this test fails to compile not because of the copyWith but because of the String passed into the constructor I think

TimWhiting:
@rrousselGit I'm happy to help review PRs if you are busy with other things. Excited for this PR!

packages/freezed/lib/src/freezed_generator.dart
DevNico: That is correct I think there should be an additional check to see if the original type was `Object`. The early return is there because Object is the supertype of everything so all "incompatible" types would be Object in the base mixin which in my opinion wouldn't be desirable.

packages/freezed/lib/src/freezed_generator.dart
DevNico: We would keep getting the supertype with each previous supertype until we hit Object. And if Object isn't the initial type we'll return since as per my other statement I think it's not desirable to have each common parameter as Object in the base mixin.

packages/freezed/test/integration/single_class_constructor.dart
DevNico: Good catch that was the autoformatter removing the const.

rrousselGit:
@TimWhiting Sure go ahead!

I should be a bit more available by September.

packages/freezed/test/multiple_constructors_test.dart
DevNico: The test was a bad rewrite of the earlier test for the different nullability. Yes it fails due to the incorrect string constructor paramter but it would also fail for the parameter which it was intended for.

I'll fix the test and also try to find a way of generally finding common subtypes.

DevNico:
@rrousselGit all two types have `Object` / `Object?` as a common supertype (except for `dynamic` maybe? I'd have to check that) would you say it is desirable to have those parameters in the common base mixin as `Object`?

As an example take:
```dart
factory Foo.first(String a) = _First;
factory Foo.second(double a) = _Second;

Foo foo;
Object a = foo.a;
```

rrousselGit:
If all the cases possess the same variable, I think showing Object/Object? is fine.
No dynamic though, that's for sure.

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -248,26 +240,48 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
...
+            final matchingParameter = constructor.parameters.allParameters
+                .firstWhereOrNull((p) => p.name == parameter.name);
 
             if (matchingParameter == null) return null;
-            if (matchingParameter.isFinal) hasAnyFinalProperty = true;
-          }
 
-          if (hasAnyFinalProperty) {
-            return parameter.copyWith(isFinal: true);
+            final commonTypeWithMatch =
```
miDeb, line 259: So, if I'm reading this correctly, this code repeatedly checks the least upper bound for the parameter in the first constructor and the parameter in another constructor. Only if this always returns the same result it will be considered.

I think this could be improved by declaring a `commonType` outside of the `for` loop, initializing it to the first parameter type, and then in the loop for each parameter type updating the `commonType` to the least upper bound between the `commonType` so far and the current parameter type.

That way the least upper bound between all parameter types will be found, even if that is different from the least upper bound between the first constructor and the second. Consider this example:

```dart
class A {}
class B implements A {}
class C implements B {}
class D implements A {}

@freezed
class Container {
  const factory Container.one({required B param}) = _One;
  const factory Container.two({required C param}) = _Two;
  const factory Container.three({required D param}) = _Three;
}
```
The types `B`, `C` and `D` are all subclasses of `A`, but currently this is not recognized.

DevNico: Tim already pointed that out. I have added a test for your use case and the parameter is available as expected.

miDeb: Sorry, I missed that.

miDeb:
We should probably still do a comparison by the string representation of the type, to make sure this doesn't break:
```dart
@freezed
class Container with _$Container {
  const factory Container.one() = _One;
  const factory Container.two() = _Two;
  const factory Container.three() = _Three;
}

@freezed
class ContainerWrapper with _$ContainerWrapper {
  const factory ContainerWrapper.one({required _One a}) = _WOne;
  const factory ContainerWrapper.two({required _One a}) = _WTwo;
}
```

`One` is not recognized because it is not yet generated, but we should generate a common getter as it is currently the case.

miDeb:
There is also a bug when handling typedefs: Consider this example (to be placed into the integration test directory)
```dart
import 'external_typedef.dart';
import 'external_typedef_two.dart' as two;

@freezed
class ContainsTypedefs with _$ContainsTypedefs {
  const factory ContainsTypedefs.one(
      {required ExternalTypedef a, required two.ExternalTypedefTwo b}) = _TOne;
  const factory ContainsTypedefs.two(
      {required ExternalTypedef? a, required two.ExternalTypedefTwo? b}) = _TTwo;
}
```

The compiled code has compile errors because we apparently "forget" that this is a typedef after searching for the least upper bound.

DevNico:
I'll fix those and any other cases you might find and add the respective tests later

miDeb:
Here's another convoluted one for you :wink: 
```dart
@freezed
class CommonType with _$CommonType {
  const factory CommonType.one({required int a}) = _One;
  const factory CommonType.two({required num a}) = _Two;
}

@freezed
class CommonTypeContainer with _$CommonTypeContainer {
  const factory CommonTypeContainer({required CommonType internal}) = _CommonTypeContainer;
}

void main() {
  final container = CommonTypeContainer(
    internal: CommonType.two(a: 0.0),
  );
  container.copyWith.internal();
}
```

This fails at runtime due to the type cast inserted [here](https://github.com/rrousselGit/freezed/pull/740/files#diff-d617954163786415754a92b9e87692ac0567ac0b028230ea42bb3246231370d4R208). If you remove it, the compiler will tell you that something is wrong: Because the common `copyWith` is generated using a common subtype, calling it with anything other than this common subtype (or _its_ subtypes) is not allowed. However, some union variants might contain a value that is not this common subtype, and if you don't specify an argument for you `copyWith` call this value will be used as the default value, which will then fail.

rrousselGit:
CopyWith should not include properties that have a type difference

DevNico:
> CopyWith should not include properties that have a type difference

So no common subtypes? But would you still want types of different nullability?

TimWhiting:
> CopyWith should not include properties that have a type difference

Why? 
For example in the null-case:
String?
String
common subtype == String, which is why copyWith passing a String makes sense because it works for all variations of the parameter. (The original PR had this)

For a not much more complicated hierarchy.
num
double
common subtype == double -> you can use double for the parameter, because no matter which constructor / union was used, double makes sense.

miDeb:
> Unlike a common supertype(leastUpperBound) which is used for getters, common subtypes don't end up with something general like Object?

Exactly. For common subtypes it will only consider the types that the constructors use, so if one constructor uses a type that is a subtype of all the types the other constructors use, this type can be used for the general copyWith implementation.

miDeb:
`unfreezed` classes don't work yet:

```dart
@unfreezed
class CommonType with _$CommonType {
  const factory CommonType.one({required int a}) = _One;
  const factory CommonType.two({required num a}) = _Two;
}
```
produces:

The return type of getter 'a' is 'num' which isn't a subtype of the type 'int' of its setter 'a'.
Try changing the types so that they are compatible.dart [getter_not_subtype_setter_types](https://dart.dev/diagnostics/getter_not_subtype_setter_types)

DevNico:
> ```dart
> @freezed
> class CommonType with _$CommonType {
>   const factory CommonType.one({required int a}) = _One;
>   const factory CommonType.two({required num a}) = _Two;
> }
> ```

That's actually quite tricky. I feel like the only solution would be to use the integrated map function and perform the `copyWith` on each concrete class. Not sure if there are any other ways of doing this.

Instead of the not compiling / running
```dart
return _then(_value.copyWith(
  a: a == freezed
      ? _value.a
      : a // ignore: cast_nullable_to_non_nullable
          as int,
));
```

We'd have something like
```dart
return _then(_value.map(
  one: (value) => value.copyWith(
    a: a == freezed
        ? value.a
        : a // ignore: cast_nullable_to_non_nullable
            as int,
  ),
  two: (value) => value.copyWith(
    a: a == freezed
        ? value.a
        : a // ignore: cast_nullable_to_non_nullable
            as num,
  ),
));
```

Do you guys have any other suggestions and or opinions on this? @TimWhiting @miDeb @rrousselGit

DevNico:
I fixed all the other cases except for the `copyWith`. For that one, I'll wait for you to give your opinion.

TimWhiting:
@DevNico 
Turns out that it has the same error for null/non-null case as well. CopyWith is very hard it seems. I think that your alternative makes sense, but I think it would probably be good to not generate the `.map` unless there is any parameter with a common subtype. Additionally, I think that this feature would be good to independently enable / disable. (Maybe call it enhancedCopyWith)?

Additionally, I keep wondering why you changed some of the lists to Iterables? Can you explain the reason for that change?

DevNico:
I changed them to Iterable since it was just unnecessary in some places to do the toList since the Iterable is sufficient. AFAIK it's a little bit more efficient not to do the toList when not needed

DevNico:
I'd really like to wait for feedback from @rrousselGit here. I do agree with you @TimWhiting that an `enhancedCopWith` would make sense. We can add all common supertypes to the base mixin by default as a new feature.

The common subtype `enhancedCopyWith` using `map` would then only be generated when this new opt-in flag is turned on _and_ a property with a common subtype exists.

There is also the option of just always generating the `map` call when there is a common subtype e.g. not having the feature be opt-in. This might have performance implications tho which would have to be evaluated whether these are negligible or not.

rrousselGit:
I don't really have time to give feedback at the moment.

But I think copyWith support would better be done in a separate PR. 
copyWith on different properties would likely involve a lot of changes to get it right. And it has a bunch of edge-cases 

Like:

```dart
@freezed
class CommonType with _$CommonType {
  const factory CommonType.one({required int a}) = _One;
  const factory CommonType.two({required num a}) = _Two;
}
```

vs:

```dart
@freezed
class CommonType with _$CommonType {
  const factory CommonType.one({required int a}) = _One;
  const factory CommonType.two({required double a}) = _Two;
}
```

or:

```dart
@freezed
class A with _$A {
  const factory A(int count) = _A;
}

@freezed
class B with _$B {
  const factory B(int count) = _B;
}

@freezed
class CommonType with _$CommonType {
  const factory CommonType.one(A value) = _One;
  const factory CommonType.two(B value) = _Two;
}

void main() {
  CommonType type;

  type = type.copyWith.value(count: 42);
}
```

packages/freezed/lib/src/templates/abstract_template.dart
```diff
@@ -13,15 +13,17 @@ class Abstract {
 
   final Data data;
   final CopyWith? copyWith;
-  final List<Property> commonProperties;
+  final Iterable<Property> commonProperties;
```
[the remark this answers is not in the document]
rrousselGit: Same for the other places where this was changed

packages/freezed/lib/src/templates/copy_with.dart
```diff
@@ -65,23 +63,26 @@ ${_abstractDeepCopyMethods().join()}
...
-              showDefaultValue: false,
-              isRequired: false,
-              defaultValueSource: '',
-              type: e.type,
-              doc: e.doc,
-              isPossiblyDartCollection: e.isPossiblyDartCollection,
-            );
-          }).toList(),
+          namedParameters: allProperties
+              .map((e) => Parameter(
```
rrousselGit, line 67: Don't use => for big function blocks like that.

{ return } is voluntary

packages/freezed/lib/src/templates/properties.dart
```diff
@@ -106,13 +136,16 @@ class Property {
...
       );
 
-  Setter get unimplementedSetter => Setter(
-        name: name,
-        type: type,
-        decorators: decorators,
-        doc: doc,
-        body: ' => throw $privConstUsedErrorVarName;',
-      );
+  Setter? get unimplementedSetter =>
```
rrousselGit, line 139: I don't like that this is nullable. It doesn't seem like the correct place to put logic here.

packages/freezed/lib/src/utils.dart
```diff
@@ -71,3 +71,16 @@ String _fixCase(String input, String separator) =>
 
       return lower;
     });
+
+String typeStringWithoutNullability(String typeString) {
```
rrousselGit, line 75: Why do we need those utilities? `getDisplayString` has a `withNullability` parameter

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,231 @@
...
+
+      value = value.copyWith(a: 'b', c: 24.42);
+      expect(value.a, 'b');
+      expect(value.c, 24.42);
+    });
+
+    test(
+      'copy has shared params with common subtypes available',
+      () async {
+        await expectLater(compile(r'''
```
rrousselGit, line 53: For new tests checking compilation errors, it'd be great to use my expect_error package instead

This  `compile`  is technical dept and should be removed at some point.

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,231 @@
...
+  CommonInterfaceSupertype param =
+    CommonInterfaceSupertype(CommonInterfaceB());
+  param.copyWith;
+}
+'''), throwsCompileError);
+    });
+  });
+
+  group('CommonTypedefs', () {
+    test('generates correct typedefs', () async {
```
rrousselGit, line 151: What is this test trying to do?

In general I don't like the idea of analyzing the source code like that. Chances are there's a better way to do this that doesn't rely on the super complex analyzer API

miDeb:
> Do you guys have any other suggestions

I could come up with another solution (not sure if we should continue this discussion in this PR though):

We could amend the concrete implementations of `freezed` classes by a new private function that allows to supply the result type and the `then` function to its `CopyWith` class. Something like this:
```dart
  @override
  @JsonKey(ignore: true)
  _$$_OneCopyWith<$Res> _$copyWith<$Res>($Res Function(_$_One) _then) =>
      __$$_OneCopyWithImpl<$Res>(this, _then);
```

Then, instead of generating a common concrete `copyWith` implementation the abstract class could return the specific concrete `CopyWith` class:

```dart
  factory $CommonTypeCopyWith(
          CommonType value, $Res Function(CommonType) then) =>
      value._$copyWith<$Res>(then);
```

packages/freezed/test/common_types_test.dart
DevNico: I literally just copied the other typedef test and adapted it here. If you don't like it like that I can rewrite it to be like the other tests

packages/freezed/lib/src/utils.dart
DevNico: That is true but for not generated types (e.g. other Freezed classes) the common type checking is done on Strings not DartTypes.

packages/freezed/lib/src/templates/abstract_template.dart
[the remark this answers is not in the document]
DevNico: I changed every output / variable back to List.

DevNico:
I removed variables where the common subtype doesn't match the common supertype e.g. it won't work without refactoring copyWith so this pr can be merged as just the common getters feature.

The improved version of the common copyWith can then be implemented in a later PR.

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -438,8 +482,8 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
   }
 
   Iterable<CloneableProperty> _commonCloneableProperties(
-    List<ConstructorDetails> constructors,
-    List<Property> commonProperties,
+    Iterable<ConstructorDetails> constructors,
```
TimWhiting, line 485: There are a couple of Iterable changes still in this PR

DevNico: I left these in because when used as a parameter there is no lazy iterable implication but giving more flexibility in the inputs to the function.

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -247,27 +229,89 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
...
+                  commonSupertypeString, matchingParameter.type!)) {
+                return null;
+              }
+
+              if (commonSupertypeDartType.isNullable !=
+                  matchingParameterType.isNullable) {
+                commonSupertypeString =
+                    typeStringWithNullability(commonSupertypeString);
+                commonSubtypeString =
+                    typeStringWithoutNullability(commonSupertypeString);
```
TimWhiting, line 284: should this be `typeStringWithoutNullability(commonSubtypeString);`

packages/freezed/lib/src/templates/copy_with.dart
```diff
@@ -197,10 +200,17 @@ $s''';
...
         propertyName = '_$propertyName';
       }
-      var ternary =
-          '${p.name} == freezed ? $accessor.$propertyName : ${p.name} ';
+
+      var ternary = '${p.name} == freezed ? $accessor.$propertyName ';
+
+      final type = p.commonSubtype ?? p.type;
+      if (p.commonSubtype != null && p.type != p.commonSubtype) {
+        ternary += 'as $type ';
```
TimWhiting, line 208: I could be wrong, but isn't this still doing the invalid casts for copyWith for a common subtype of `num, int -> int`?

Are we moving the copyWith changes to another PR or are you still trying to do the `T & T? -> T` subtyping (which doesn't run into cast errors right?)

DevNico: This is only for the nullability subtyping which shouldn't run into problems at all.

packages/freezed/lib/src/freezed_generator.dart
DevNico: I'll take a look at this again next week!

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,173 @@
...
+// import 'integration/common_types.dart';
+
+// void main() {
+//   final param = CommonSuperSubtype('a', 1337, 42);
+
+//   // expect-error: UNDEFINED_NAMED_PARAMETER
+//   param.copyWith(b: 42);
+// }
+// '''), compiles);
+//     });
```
rrousselGit, line 71:
```suggestion
```

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,173 @@
+import 'package:analyzer/dart/analysis/results.dart';
+import 'package:analyzer/dart/element/element.dart';
+import 'package:analyzer/dart/element/type.dart';
+import 'package:build_test/build_test.dart';
+import 'package:expect_error/expect_error.dart';
+import 'package:test/test.dart';
+
+import 'integration/common_types.dart';
+
+// TODO: Uncomment tests when copyWith for common subtypes is implemented
```
rrousselGit, line 10:
```suggestion
```

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,143 @@
...
+  //       );
+  //       commonTypeContainer = commonTypeContainer.copyWith.internal();
+  //       expect(commonTypeContainer.internal.a, 0.0);
+
+  //       commonTypeContainer = commonTypeContainer.copyWith.internal(a: 12);
+  //       expect(commonTypeContainer.internal.a, 12);
+  //     },
+  //   );
+  // });
+
```
rrousselGit, line 118:
```suggestion
```

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,122 @@
...
+            .getErrors('/freezed/test/integration/common_types.freezed.dart')
+        as ErrorsResult;
+
+    expect(errorResult.errors, isEmpty);
+  });
+
+  group('CommonSuperSubtype', () {
+    test('has the common properties available', () {
+      var value = const CommonSuperSubtype('a', 1337, 42.24);
+      expect(value.a, 'a');
```
rrousselGit, line 32: This should test that the public type of those properties is respectively `String?`, `num` and `num`

Technically this test could expose those properties as `Object?`, and the test would still pass

packages/freezed/test/common_types_test.dart
```diff
@@ -0,0 +1,122 @@
...
+      var e = freezedClass.getGetter('genericTypedef')!.returnType;
+      expect(e, isA<FunctionType>());
+      expect(e.alias!.element.name, equals('GenericTypedef'));
+      expect(e.alias!.typeArguments.toString(), equals('[int, bool]'));
+    });
+  });
+
+  group('CommonUnfreezed', () {
+    test('should have common getters available', () {
+      var value = CommonUnfreezed.one(a: 42, b: 3.14);
```
rrousselGit, line 105: Same as the other test. We want to make sure that shared getting are correctly typed.

rrousselGit:
There are a few conflicts and some small tests changes. But otherwise LGTM!

Thanks for your work 😄

rrousselGit (approved):

rrousselGit:
Looking back at the logic for obtaining the common subtype between union cases, the logic seems to do more than necessary

Typedefs and generated types are not supported by Freezed in general. Trying to support it for this specific case doesn't make sense, as they are still unsupported for everything else.

rrousselGit:
Alright, time to merge this!

Sorry for the delay :)

PR #742, opened 2022-08-15
fix: use nullable return type in *OrNull callbacks

### LINKED ISSUE #741: mapOrNull callback return type is non-null
Currently the generated `mapOrNull` / `whenOrNull` callbacks specify a non-null return type.

```dart
@optionalTypeArgs
TResult? mapOrNull<TResult extends Object?>({
  TResult Function(A value)? a,
  TResult Function(B value)? b,
})
```

This forces us to write something like
```dart
value.mapOrNull<int?>(a: (v) => v.abc));
```
when we access a nullable property.

DevNico:
closes #741


rrousselGit:
Hello!

Sorry for the delay. Could you write tests for this?

DevNico:
Will do next week!

DevNico:
Done :)

rrousselGit:
Thanks a lot!

PR #759, opened 2022-09-13
Update pub.dev links

nohli:
With this PR, the right section on pub.dev will show the View/report issues link to GitHub again, like [here](https://pub.dev/packages/android_id) (the show-logic was recently changed).

FYI with the issue_tracker link, you could also directly select issue labels (like `freezed` or `freezed_annotation`) when opening the link, if desired, like [here](https://pub.dev/packages/device_info_plus).


rrousselGit:
Thanks!

PR #771, opened 2022-10-03
fix(freezed): fix isPossiblyDartCollection to not return true for any type

### LINKED ISSUE #770: isPossiblyDartCollection returns true for any type
**Describe the bug**

The current implementation of `isPossiblyDartCollection` returns always true, even for types that are definitely no collection, like `int`.

```dart
  bool get isPossiblyDartCollection {
    final interface = safeCast<InterfaceType>();

    return isDartCoreMap ||
        isDartCoreIterable ||
        isDartCoreSet ||
        isDartCoreList ||
        isDynamic ||
        isDartCoreObject ||
        this is TypeParameterType ||
        (interface != null &&
            interface.allSupertypes.any((e) => e.isPossiblyDartCollection));
  }
```

The bug is in the recursion -> all types have `Object` as superType, and so `isDartCoreObject` will return true for the supertype, and so `interface.allSupertypes.any((e) => e.isPossiblyDartCollection)` will always return true.

Due to this, the equality check will always use `DeepCollectionEquality().equals()`, which is okay from a correctness perspective, but will impact performance negatively. 

**To Reproduce**

Kind of hard to reproduce this, since we need instances of `DartType`. But it's easy to reproduce the bug by looking at the result of code generation.

See for example the `==` method on Appointement:

```dart
  @override
  bool operator ==(dynamic other) {
    return identical(this, other) ||
        (other.runtimeType == runtimeType &&
            other is _$_Appointment &&
            const DeepCollectionEquality().equals(other.title, title) &&
            const DeepCollectionEquality().equals(other.date, date) &&
            const DeepCollectionEquality().equals(other.creator, creator));
  }
```
This calls DeepCollectionEquality for `title` even though it's a String. 

if we remove the recursive `isPossiblyDartCollection` check (which is obviously not to correct solution, as it would not detect Uint8List etc. correctly) we get this result:

```dart
  @override
  bool operator ==(dynamic other) {
    return identical(this, other) ||
        (other.runtimeType == runtimeType &&
            other is _$_Appointment &&
            (identical(other.title, title) || other.title == title) &&
            (identical(other.date, date) || other.date == date) &&
            (identical(other.creator, creator) || other.creator == creator));
}
```

**Expected behavior**

For basic types like String, int, num, `DeepCollectionEquality` should not be used instead of the `==/identical` check

knaeckeKami:
fixes #770 

I would like to add tests for this, but as this method acts directly on DartType, which, to my knowledge, cannot be instantiated in a meaningful way directly,  I omitted tests for now.


rrousselGit:
That's fine. It counts more as a performance optimization than a behavior change. A benchmark would be what tests it.

PR #772, opened 2022-10-05
Allow analyzer < 5.0.0

joeldomke:
Forcing analyzer to  ^5.0.0 leads to conflicts with other packages, e.g. intl_utils

### DIFF
--- a/packages/freezed/pubspec.yaml
+++ b/packages/freezed/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   sdk: '>=2.17.0 <3.0.0'
 
 dependencies:
-  analyzer: ^5.0.0
+  analyzer: ">=4.2.0 <6.0.0"
   build: ^2.0.0
   build_config: ^1.0.0
   collection: ^1.15.0


rrousselGit:
It probably makes more sense to ask those packages to upgrade instead

5.0.0 has breaking changes. It's not backward compatible with 4.0.0

SunlightBro:
Just tested [@kevmoo](https://twitter.com/kevmoo/status/1577752391012986880)'s idea
`dart pub downgrade analyzer --dry-run`
```sh
< _fe_analyzer_shared 48.0.0 (was 49.0.0)
< analyzer 5.0.0 (was 5.1.0)
Would change 2 dependencies.
```

**BUT** `dart pub downgrade` does not support `--major-versions` like it's (`dart pub upgrade`) counter-part ...

rrousselGit:
I don't want to do it on a per package basis. I'd like that command to work without specifying `analyzer` and apply to all packages at once.

rrousselGit:
Not sure how I missed the command. But hey :shrug: 
I don't want to specify `analyzer` though. Just "dart pub downgrade" is what I want.

Fancy updating the CI to have a step do that?

packages/freezed/pubspec.yaml
```diff
@@ -10,7 +10,7 @@ environment:
   sdk: '>=2.17.0 <3.0.0'
 
 dependencies:
-  analyzer: ^5.0.0
+  analyzer: ">=4.2.0 <6.0.0"
```
SunlightBro, line 13: Did some extensive tests, and the lowest constraint, where the test still succeed is:
`analyzer: ">=4.6.0 <6.0.0"`

SunlightBro:
@rrousselGit Always run those downgraded test , or only conditionally when the pubspec.yaml has changes ?

rrousselGit:
Neither

It's about having a matrix, where the CI tests different environments: lowest version vs highest version

joeldomke:
Thanks!

PR #776, opened 2022-10-06
Test pub downgrade via matrix

SunlightBro:
Hey @rrousselGit , this PR is in regard of your [comment](https://github.com/rrousselGit/freezed/pull/772#issuecomment-1269683090) on #772 .

I did a lot a lot of testing of all possible combinations of version-constraints, so that all tests still succeed.
Also always cross-checking locally with `dart pub deps` which transitive dependencies make the tests fail when they downgrade to "low".
You can have a look at the [actions on my fork](https://github.com/SunlightBro/freezed/actions/workflows/build.yml)

I had to introduce a dependency_override `file` 😞 as build_runner will fail below `file: ^6.1.3`
```yaml
  # pub downgrade:
  # build_runner 2.2.0 => glob 2.0.0 => file 6.0.0
  # but we need minimum 6.1.3
  file: ^6.1.3
```

For now I did not do any changes to the `_internal` package , cause I don't think we have to.

Let me know if want some more comments/explanations on my changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,7 @@
 name: Build
 
 on:
+  workflow_dispatch:
   push:
   pull_request:
   schedule:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,6 +24,9 @@ jobs:
         dependencies:
           - get
           - downgrade
+        exclude:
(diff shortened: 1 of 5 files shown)


rrousselGit:
That's some epic work right there. Will definitely help a ton

PR #777, opened 2022-10-07
Cache flutter sdk

SunlightBro:
Added cache to the `subosito/flutter-action@v2` step.
Only on `stable` though, because on master the **cache-key** always looked the same.

- `stable`cache-key `flutter-linux-stable-3.3.4-x64-eb6d86ee27deecba4a83536aa20f366a6044895c`
- `master`cache-key `flutter-linux-master-master-x64-master`

Also updated all 'uses' steps to their latest version

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-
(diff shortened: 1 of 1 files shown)


rrousselGit:
Thanks a lot!

Issue #833, opened 2023-01-05
fields contains `$` failed!


medz:
**Describe the bug**
Including `$` strings in classes fields will cause dart run build_runner to fail.

<img width="609" alt="image" src="https://user-images.githubusercontent.com/5564821/210706078-a9b6a7bf-efa6-4a27-9fe7-d9d197b7d217.png">

medz:
<img width="780" alt="image" src="https://user-images.githubusercontent.com/5564821/210706128-51e67eb5-6d12-431b-a175-890cd972cfe4.png">

SunlightBro:
this is not a freezed issue.
You either need to escape the $ character `'\$'`
or use a raw String `r'$'`

medz:
> Fancy fixing it?

I need to try it when I have some free time.

> I hate the Dart Build System, but I have to use it now.

medz:
@rrousselGit The PR fixed:

- #835

PR #835, opened 2023-01-08
Support class name/named constructor/class fields contain $ symbols

medz:
- [x]  #833

### DIFF
--- a/packages/freezed/lib/src/templates/prototypes.dart
+++ b/packages/freezed/lib/src/templates/prototypes.dart
@@ -33,10 +33,6 @@ String encodeClassOrField(String name) {
     buffer.writeCharCode(char);
   }
 
-  if (name.contains('\$')) {
-    print(buffer.toString());
-  }
-
   return buffer.toString();
 }
 
--- a/packages/freezed/lib/src/templates/prototypes.dart
+++ b/packages/freezed/lib/src/templates/prototypes.dart
@@ -12,6 +12,32 @@ List<String> parseDecorators(List<ElementAnnotation> metadata) {
   ];
 }
 
+String variable(String name) {
+  return name.contains(r'$') ? '\${$name}' : '\$$name';
+}
+
+String encodeVariable(String name) {
+  final buffer = StringBuffer();
+
+  for (final char in name.codeUnits) {
+    if (char == r'$'.codeUnitAt(0)) {
+      // Whether the previous character was a `\` and if so skip it.
+      if (buffer.isNotEmpty &&
+          buffer.toString().substring(buffer.length - 1) == r'\') {
+        buffer.write(char);
+        continue;
+      }
+
+      buffer.write(r'\$');
+      continue;
+    }
+
+    buffer.writeCharCode(char);
+  }
+
+  return buffer.toString();
+}
+
 extension FreezedElementAnnotation on ElementAnnotation {
   /// if the element is decorated with `@Default(value)`
   bool get isDefault {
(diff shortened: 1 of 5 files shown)


Woutwo:
Awesome, just ran in this issue when using generated classes by the graphql codegen

packages/freezed/test/integration/concrete.dart
```diff
@@ -96,3 +96,10 @@ class Const with _$Const {
   const Const._();
   const factory Const(int a) = _Const;
 }
+
+@freezed
+class Dollar with _$Dollar {
```
rrousselGit, line 101: We'd want to test $ in the middle/end of the variable name too.

And in constructor names

packages/freezed/lib/src/templates/prototypes.dart
```diff
@@ -12,6 +12,32 @@ List<String> parseDecorators(List<ElementAnnotation> metadata) {
   ];
 }
 
+String variable(String name) {
+  return name.contains(r'$') ? '\${$name}' : '\$$name';
+}
+
+String encodeVariable(String name) {
+  final buffer = StringBuffer();
```
rrousselGit, line 20: Is there a reason why we can't do `return name.replaceAll(r'$' ,r'\$')`?

packages/freezed/test/integration/concrete.dart
medz: > We'd want to test $ in the middle/end of the variable name too.
> 
> And in constructor names

@rrousselGit Added, now both named constructors and class names support interspersed $ characters

packages/freezed/lib/src/templates/prototypes.dart
medz: Of course there is! If the input string is "a\$$$\$b\$" you? Or accidentally call the wrapper repeatedly! I changed the function name (accurately expressing the intent of the function) in the latest code to be compatible with multiple calls.

medz:
@rrousselGit You need to review the new code.

packages/freezed/lib/src/templates/prototypes.dart
rrousselGit: Why would the input contain `\$` in it? I don't think that's a valid character for a property name, right?

rrousselGit:
Some tests seem to be failing. Mind having a look?

medz:
> Some tests seem to be failing. Mind having a look?

My mistake, now all tests are successful

andreasgan:
Anything preventing a merge here @rrousselGit?

packages/freezed/test/concrete_test.dart
```diff
@@ -34,5 +34,12 @@ void main() {
     expect(CustomToString().toString(), '42');
     expect(MixedInToString().toString(), '42');
     expect(BaseToString().toString(), '42');
+    expect(const Dollar($test: 'demo').toString(), r'Dollar($test: demo)');
+    expect(const DollarMiddle(te$st: 'demo').toString(),
+        r'DollarMiddle(te$st: demo)');
+    expect(
+        const DollarEnd(test$: 'demo').toString(), r'DollarEnd(test$: demo)');
```
rrousselGit, line 41:
```suggestion
    expect(
      const DollarMiddle(te$st: 'demo').toString(),
      r'DollarMiddle(te$st: demo)',
    );
    expect(
      const DollarEnd(test$: 'demo').toString(),
      r'DollarEnd(test$: demo)',
    );
```

rrousselGit:
Moving to a separate branch because Github is confused since the original fork has been deleted

Issue #254, opened 2020-09-03
Add field name strings as static members of class to avoid typing them manually when needed


dannnnthemannnn:
https://gist.github.com/dannnnthemannnn/f22b94c586c26a0909be073d2cceed42

`
@freezed
abstract class Discussion with _$Discussion {
  factory Discussion({
    DateTime scheduledTime,
    bool isPublic,
    List<String> participantIds,
  }) = _Discussion;
  
  // Add Helper field name somehow to generated class similar to this:
  static const scheduledTimeField = 'scheduledTime';
  static const isPublicField = 'isPublic';
  static const participantIds = 'participantIds';
}
`

rrousselGit:
This could be added to the generated `$Discussion` class, such that we have:

```dart
print($Dicussion.nameOf.scheduledTime) // scheduledTime
```

mono0926:
I want this feature too 👀

dannnnthemannnn:
Would this make more sense on json_serializable as this is probably most used with json serialization/deserialization?

deimantasa:
Would love to have it as well.

esenmx:
> This could be added to the generated `$Discussion` class, such that we have:
> 
> ```dart
> print($Dicussion.nameOf.scheduledTime) // scheduledTime
> ```

An alternative name would be `Descriptor`, similar to protobuf.

jan-siroky:
Hey, this would be great. Is there any way I could help? Beside PR? I feel that it would take too much time for me to dive into your codebase.

I am certainly going to donate once this is the feature in freezed package :-)

Thanks for consideration!

rrousselGit:
I  have no plan to implement this, so closing.

IMO this is not the responsibility of Freezed, and should instead be done by a separate code-generator.

PR #879, opened 2023-03-23
feat: Add NameOf class to lookup property names as string constants

MBulli:
Hi @rrousselGit,  I suggest making the following changes to implement #254.

You can use it like this:

main.dart:
```dart
@freezed
class MyClass with _$MyClass {
  factory MyClass({String? a, int? b}) = _MyClass;
}

void main() {
  // nameOf
  print(_$MyClass.nameOf.a);
  print(_$MyClass.nameOf.b);
}
```

main.freezed.dart:
```dart
/// @nodoc
class _$MyClassNameOf {
  String get a => "a";
  String get b => "b";
}

/// @nodoc
mixin _$MyClass {
  String? get a => throw _privateConstructorUsedError;
  int? get b => throw _privateConstructorUsedError;
  // ignore: unused_field
  static late final _$MyClassNameOf nameOf = _$MyClassNameOf();

  @JsonKey(ignore: true)
  $MyClassCopyWith<MyClass> get copyWith => throw _privateConstructorUsedError;
}
```

There is an option to enable/disable the code generation of the NameOf class. Since the issue has been open for a while, it appears that there is not much demand for this feature. Therefore, I have decided to disable the feature by default.

Please let me know your thoughts on this. 
If you are open to the feature and the proposed implementation, I will also add a test case.

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -110,6 +110,7 @@ class FreezedGenerator extends ParserGenerator<GlobalData, Data, Freezed> {
       generateEqual: configs.equal ?? !_hasCustomEquals(element),
       generateFromJson: configs.fromJson ?? await needsJsonSerializable,
       generateToJson: configs.toJson ?? await needsJsonSerializable,
+      generateNameOf: configs.nameOf!,
       genericArgumentFactories: configs.genericArgumentFactories,
       map: MapConfig(
         map: configs.map?.map ?? shouldGenerateUnions,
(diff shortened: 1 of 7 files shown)


rrousselGit:
Honestly I'm not fan of adding this.

MBulli:
Do you dislike the feature or the implementation?

MBulli:
If you don't like this feature in your library, I'll close this PR and you can close the issue imho.

I would actually use it in my project, but I'll see what I can achieve with another pub or maybe my own builder. Still, it was a nice exercise for me.

PR #903, opened 2023-05-10
fix `DartType.isDynamic` deprecation 

SunlightBro:
With analyzer 5.12.0 
`DartType.isDynamic` got deprecated,
instead use `this is DynamicType`.

### DIFF
--- a/packages/freezed/lib/src/freezed_generator.dart
+++ b/packages/freezed/lib/src/freezed_generator.dart
@@ -167,7 +167,7 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
       if (parameter.type.nullabilitySuffix != NullabilitySuffix.question &&
           parameter.isOptional &&
           parameter.defaultValue == null &&
-          !(parameter.type is DynamicType)) {
+          parameter.type is! DynamicType) {
         final ctorName =
             constructor.isDefaultConstructor ? '' : '.${constructor.name}';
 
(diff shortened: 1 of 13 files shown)


rrousselGit:
That's a wild change IMO, but sure. Will review a bit later :)

rrousselGit:
Yeah I was reading that.

I would assume that yes it is impacted. But I'll have to double check when I'll give this a proper review

SunlightBro:
You can revert [`8490802`](https://github.com/rrousselGit/freezed/pull/903/commits/84908024240e74efcaded0602c90744537340b99) if it's not something you want.
I just wanted to see if it would run the Checks successfully.

rrousselGit:
Thanks for fixing the CI~

packages/freezed/lib/src/freezed_generator.dart
```diff
@@ -167,7 +167,7 @@ Read here: https://github.com/rrousselGit/freezed/blob/master/packages/freezed/C
       if (parameter.type.nullabilitySuffix != NullabilitySuffix.question &&
           parameter.isOptional &&
           parameter.defaultValue == null &&
-          !parameter.type.isDynamic) {
+          !(parameter.type is DynamicType)) {
```
rrousselGit, line 170:
```suggestion
          parameter.type is! DynamicType) {
```

rrousselGit:
I think InvalidType should be checked on all of these cases too.

One of the issues is what could happen when the output of a code generator is the input of a Freezed class.  
We'd need to test something that for all of these `is DynamicType` cases

For example for the "is possibly a collection", someone could do:

```dart
@freezed
  factory Example(
    GeneratedClassWhichImplementsList list,
  );
}
```

where `GeneratedClassWhichImplementsList` is not available at the moment where Freezed runs.

rrousselGit:
Before there was no distinction between dynamic and invalid, so the tests simply used dynamic to mimic an invalid type. But that's not the case anymore so we'd probably have to duplicate the tests for those.

Issue #916, opened 2023-05-20
Incompatibility with dart 3


josiahsrc:
**Describe the bug**
Starting from dart Dart 3.0.1 (shipped with flutter `3.10.1`), nested classes produce an `InvalidType` when generated. E.g. given this code

```dart
@freezed
class SinglePage with _$SinglePage {
  factory SinglePage(Page page) = _SinglePage;
}

@freezed
class WidgetType with _$WidgetType {
  const factory WidgetType.page() = Page;
}
```

Freezed produces this (as one example of the `InvalidType` occurrence, there are others.

```dart
...
mixin _$SinglePage {
  InvalidType get page => throw _privateConstructorUsedError;
  ...
}
```

**To Reproduce**

1. Update your dart version to 3.0.1

2. Create a dart project with freezed installed

3. Paste this content into a `lib/test.dart` file
```dart
import 'package:freezed_annotation/freezed_annotation.dart';

part 'test.freezed.dart';

@freezed
class SinglePage with _$SinglePage {
  factory SinglePage(Page page) = _SinglePage;
}

@freezed
class WidgetType with _$WidgetType {
  const factory WidgetType.page() = Page;
}
```

4. Generate code

This should generate a file with `InvalidType` in place of where a nested `Page` type should be in the freezed file.

**Expected behavior**
The generated code should use `Page` instead of `InvalidType`.

josiahsrc:
I've tried debugging this locally, and it looks like this might be a problem with build-runner itself. When adding this print statement to the `generateForData` function, the build runner returns `InvalidType` for any nested freezed types. It looks like this package doesn't have control over this.

```dart
@override
  Iterable<Object> generateForData(
    GlobalData globalData,
    Data data,
  ) sync* {
    try {
      final param = data.constructors.first.parameters.allParameters.first;
      print(param.type);
    } catch (e) {
      print(e);
    }
    ...
```

I'm not sure how best to articulate this to the build runner team outside of the context of the freezed package. Any ideas?

rrousselGit:
That's not a new thing in itself.

It's weird that you get InvalidType though. I'll look into that. Normally you'd get dynamic

rrousselGit:
You cannot rely on generates types as input of other generated types in general.

rrousselGit:
Cc @SunlightBro

It may have made sense to throw on invalid type now.

Carapacik:
Same problem. I found it after updating analyzer to 5.13.0

[...]

PR #917, opened 2023-05-20
Fix incompatibility with dart 3

josiahsrc:
This is a very crude attempt at fixing an `InvalidType` code generation issue. I've isolated the problem to these two spots in the code base.

More info [here](https://github.com/rrousselGit/freezed/issues/916)

### DIFF
--- a/packages/freezed/lib/src/tools/type.dart
+++ b/packages/freezed/lib/src/tools/type.dart
@@ -73,6 +73,10 @@ String resolveFullTypeStringFrom(
     }
   }
 
+  if (displayType.contains('InvalidType')) {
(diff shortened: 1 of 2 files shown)


packages/freezed/lib/src/tools/type.dart
```diff
@@ -73,6 +73,10 @@ String resolveFullTypeStringFrom(
     }
   }
 
+  if (displayType.contains('InvalidType')) {
+    displayType = displayType.replaceAll('InvalidType', 'dynamic');
+  }
```
josiahsrc, line 78: There must be a better way to do this. Rather than matching for the `InvalidType` string, it would be better to check that the type is an `InvalidType` dart type

PR #921, opened 2023-05-23
fix regression in analyzer 5.13.0

SunlightBro:
This is only a temporary solution,

Should analyzer support something like `UnimplementedType` in the future, this needs to be reworked.

### DIFF
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -565,7 +565,7 @@ String? parseTypeSource(VariableElement element) {
     final source =
         element.source!.contents.data.substring(0, element.nameOffset);
     final eleType = element.type;
-    if (eleType is DynamicType || eleType is InvalidType) {
+    if (eleType.isDynamic2) {
       final match = RegExp(r'([$\w]+\??)\s+$').firstMatch(source);
       return match?.group(1);
     } else if (eleType is InterfaceType) {
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -560,11 +560,12 @@ extension DefaultValue on ParameterElement {
 String? parseTypeSource(VariableElement element) {
   String? type = element.type.getDisplayString(withNullability: true);
 
-  if (type.contains('dynamic') && element.nameOffset > 0) {
+  if ((type.contains('dynamic') || type.contains('InvalidType')) &&
+      element.nameOffset > 0) {
     final source =
         element.source!.contents.data.substring(0, element.nameOffset);
     final eleType = element.type;
-    if (eleType is DynamicType) {
+    if (eleType is DynamicType || eleType is InvalidType) {
       final match = RegExp(r'([$\w]+\??)\s+$').firstMatch(source);
       return match?.group(1);
     } else if (eleType is InterfaceType) {
--- a/packages/freezed/lib/src/tools/type.dart
+++ b/packages/freezed/lib/src/tools/type.dart
@@ -1,13 +1,6 @@
 import 'package:analyzer/dart/element/element.dart';
 import 'package:analyzer/dart/element/nullability_suffix.dart';
 import 'package:analyzer/dart/element/type.dart';
-import 'package:analyzer/dart/element/type_provider.dart';
-import 'package:analyzer/dart/element/type_system.dart';
-
-// DynamicTypeImpl is imported directly from source as it's not exposed by analyzer
-// ignore: implementation_imports
-import 'package:analyzer/src/dart/element/type.dart'
-    show DynamicTypeImpl, InvalidTypeImpl;
 import 'package:collection/collection.dart';
 
 import 'imports.dart';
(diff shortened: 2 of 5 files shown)


packages/freezed/lib/src/tools/type.dart
```diff
@@ -73,9 +78,41 @@ String resolveFullTypeStringFrom(
...
+  //
+  // it would generate:  List<InvalidType>
+  // instead of          List<dynamic>
+  //
+  // This a regression in analyzer 5.13.0
+  if (type is InterfaceType &&
+      type.typeArguments.contains(InvalidTypeImpl.instance)) {
+    var modified = type;
+    modified.typeArguments
+      ..replaceWhere(
```
rrousselGit, line 95: That's mutating an object which is not owned by Freezed. Could you use another approach?

packages/freezed/lib/src/tools/type.dart
```diff
@@ -73,9 +78,41 @@ String resolveFullTypeStringFrom(
...
+        (t) => t is InvalidType,
+        DynamicTypeImpl.instance,
```
rrousselGit, line 97: Use TypeSystem/TypeProvider. You should be able to obtain a dynamicType more officially

packages/freezed/test/integration/multiple_constructors.dart
```diff
@@ -165,3 +165,18 @@ class NestedMapItem with _$NestedMapItem {
   factory NestedMapItem.inner(Map<String, LeafNestedMapItem> children) =
       InnerNestedMapItem;
 }
+
+@freezed
+class ToBeGenerated with _$ToBeGenerated {
+  const factory ToBeGenerated.generated() = CodeGenerated;
```
rrousselGit, line 171: Ideally that `CodeGenerated` should not be a freezed class.

Freezed classes could be special cases, as Freezed could explicitly support those since it should know they are going to be generated

packages/freezed/test/multiple_constructors_test.dart
```diff
@@ -592,4 +592,34 @@ void main() {
           'Map<String, LeafNestedMapItem>');
     });
   });
+  group('GeneratedCode', () {
```
rrousselGit, line 595:
```suggestion

  group('GeneratedCode', () {
```

packages/freezed/test/multiple_constructors_test.dart
```diff
@@ -592,4 +592,35 @@ void main() {
...
+    test('generates Map of correct type', () async {
+      final nestedMapClass = _getClassElement('_UsesGenerated');
+
+      expect(
+          nestedMapClass
+              .getField('value')!
+              .type
+              .getDisplayString(withNullability: true),
+          'CodeGenerated');
```
rrousselGit, line 605: Use trailling commas everywhere

packages/freezed/lib/src/templates/concrete_template.dart
```diff
@@ -560,11 +560,12 @@ extension DefaultValue on ParameterElement {
...
   String? type = element.type.getDisplayString(withNullability: true);
 
-  if (type.contains('dynamic') && element.nameOffset > 0) {
+  if ((type.contains('dynamic') || type.contains('InvalidType')) &&
+      element.nameOffset > 0) {
     final source =
         element.source!.contents.data.substring(0, element.nameOffset);
     final eleType = element.type;
-    if (eleType is DynamicType) {
+    if (eleType is DynamicType || eleType is InvalidType) {
```
rrousselGit, line 568: I previously made a .isDynamic2 helper which checks both DynamicTYpe and InvalidType.

Could you use it here?

packages/freezed/lib/src/tools/type.dart
```diff
@@ -73,9 +78,41 @@ String resolveFullTypeStringFrom(
...
+  //
+  // For example for:
+  // List<ToBeGenerated> values,
+  //
+  // it would generate:  List<InvalidType>
+  // instead of          List<dynamic>
+  //
+  // This a regression in analyzer 5.13.0
+  if (type is InterfaceType &&
+      type.typeArguments.contains(InvalidTypeImpl.instance)) {
```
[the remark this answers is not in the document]
rrousselGit: We could do `any((e) => e is InvalidType)`

rrousselGit, line 92: That logic needs to be recursive. The type argument could have type arguments too.

We probably should define a separate "typeToString" function which we reuse everywhere

packages/freezed/lib/src/tools/type.dart
SunlightBro: How would I get access to the TypeProvider?
Might need an example, do you use in riverpod_generator?

rrousselGit: From LibraryElement, which is accessible from any element.

rrousselGit:
If you don't mind, I'll quickly investigate how long it would take to refactor Freezed to use Ast instead.

Ultimately that's the real solution and would fix many other issues.

rrousselGit:
I have a promising solution, but it'll take a bit of time. I'll merge your solution your now, and take my time to do the fix

Issue #610, opened 2022-03-02
Custom line-length for generated code


fzyzcjy:
Hi thanks for the wonderful lib! I would appreciate it if the generated `.g.dart` code could be 120-character in width. I can make a PR for this.

How to implement this: https://github.com/dart-lang/build/issues/3250

rrousselGit:
I don't really like this. IMO the generated code should not have to respect the style of the project in which it is generated in, as the generated code shouldn't be visible to begin with.

fzyzcjy:
Hi thanks for the reply. My use case is: I will run `dart format --line-length 120` to my *whole* project regularly (both in CI and when developing), and use that as code formatting. Could you please provide some suggestions?

rrousselGit:
Running it in dev shouldn't have any impact

And the CI shouldn't verify generated files. Maybe consider deleting generated files before checking their format

fzyzcjy:
> And the CI shouldn't verify generated files. Maybe consider deleting generated files before checking their format

Good suggestion

> Running it in dev shouldn't have any impact

Seems that all .g.dart and .freezed.dart are *reformatted* from the original 80-length to 120-length :(

rrousselGit:
> Seems that all .g.dart and .freezed.dart are reformatted from the original 80-length to 120-length :(


To begin with, you probably want the dev environment to only format modified files instead of the whole project

fzyzcjy:
Sounds reasonable, but those generated files are changed (though automatically but codegen) as well

rrousselGit:
I personally don't plan on working on this.

If you really want this, I'd be happy to review a PR. Otherwise can we close this? :smile:

ekuleshov-idexx:
The standard `dart format` command accepts list of files/folders for reformatting and it allows to fail build when project's formatting doesn't match the standard one. Unfortunately the `dart format` does not have an option to exclude generated files.

So, currently we have to reformat all generated files manually in order to pass our build.

Please reconsider adding an option to specify line length for the generated code and maybe also use the `DartFormatter` from `dart_style` to format generated code.

PR #935, opened 2023-06-10
Allow configuring line-length

TimWhiting:


rrousselGit:
I'm not sure why we care about line length in generated code.

TimWhiting:
[#610 ](https://github.com/rrousselGit/freezed/issues/610#issuecomment-1485475970)

Running `dart format --line-length 120 --exit-if-changed .` on a whole project that is configured to a line length of 120 causes issues. There are workarounds, but they are unsatisfactory. It means that there is a constant need to reformat files prior to committing if you commit generated files. Also part of the reason for wanting a longer line length in general is to make code easier to read. I find that I end up looking in generated files from time to time, and having a longer line length makes it easier to understand.

In the linked issue issue you said that although you don't plan on doing this, you'd be happy to review a PR.

In general I think it is useful, and is not much code to maintain - very little actually.

rrousselGit:
How would that work for ".g.dart" files?

I would assume that instead of `dart format . `, you should be able to use something of the form of `**/!*.freezed.dart`. Can't remember the exact syntax

TimWhiting:
.g files could be done by formatting each piece of the .g file separately, which would be not ideal, or to add a formatting option for source_gen's combiningBuilder, and have it as a separate option in the build.yaml.

rrousselGit:
Source_gen has such an option as of today?

Because if it doesn't, I'm not sure it'd made sense to offer one for Freezed.
But if it does, I don't have a problem with it

TimWhiting:
I'd be happy to contribute to source_gen. 

I'm mainly exploring alternative options to: https://github.com/dart-lang/dart_style/issues/918#, since there has been not much progress for the last few years on that issue. People care about this, and I don't see any reason not to let people change these options to their preferences, especially with how little code this is for freezed to maintain. I think you underestimate how many people will use it even if source_gen doesn't have the option. 

If the dart_style issues makes progress and adds a configuration option / file, then source_gen / build would have to be updated anyways to take such configuration into account. 

I don't see source_gen not having an option today as a reason to dismiss this. If no one starts providing an option, then it will never happen because no one wants to be the first.

There is a precedence for having this option: https://pub.dev/packages/stormberry which is a generator for postgres schemas has a line_length option.

rrousselGit:
The fact of the matter is that Freezed generates both a .g.dart and .freezed.dart

So formatting only one of them won't quite be enough

rrousselGit:
Overall I'm not against this. But I think it's important for this to work in .g files due to how Freezed is closely tied to them.

If Freezed formats .freezed, but the `dart format` command still fails due to json_serializable & stuff, the problem isn't really solved.

rrousselGit:
@TimWhiting Are you looking into making a similar PR for source_gen for the .g.dart files?

If not, should we close this?

TimWhiting:
We can close this, I don't have the time for this right now

PR #966, opened 2023-08-17
Fix watcher 1.0.0 downgrade issue (#9)

SunlightBro:
* pin watcher 1.1.0


rrousselGit:
Awesome, thanks for the fix!

PR #978, opened 2023-09-11
Remove deprecated build example

kalafut:
The `flutter pub run` version results in:

`Deprecated. Use "dart run" instead.`


rrousselGit:
Thanks

PR #997, opened 2023-10-17
Replace "dart pub add --dev" with "dev:"

tatsuyafujisaki:
This pull request is to replace `dart pub add --dev` with `dart pub add dev:` as this is the newer way.
https://dart.dev/tools/pub/cmd/pub-add#dev-dependency

Also, this pull request replaces `flutter pub add —dev` with `flutter pub add dev:` as used in the Flutter documentation.
https://docs.flutter.dev/data-and-backend/serialization/json#setting-up-json_serializable-in-a-project

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -74,22 +74,22 @@ For a Flutter project:
 
 ```console
 flutter pub add freezed_annotation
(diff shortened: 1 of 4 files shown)


rrousselGit:
Thanks!

Issue #1016, opened 2023-12-07
Generated freezed classes expand == operator parameter to be nullable (dynamic)


srawlins:
Generated freezed classes declare an `operator ==` override which expands the parameter type (from `Object`, on `Object.==`) to a nullable type (`dynamic`). For example:

```dart
  @override
  bool operator ==(dynamic other) {
    return identical(this, other) ||
        (other.runtimeType == runtimeType &&
            other is ListIndexPath &&
            const DeepCollectionEquality().equals(other.index, index));
  }
```

No Dart runtime passes a `null` value to an `==` implementation. Explicitly expanding the parameter type does not seem to serve a purpose, and can lead to confusion that the function will ever see a `null` argument value. These implementation should not have an expanded parameter type of `dynamic`.

**To Reproduce**

Generate a freezed class.

**Expected behavior**

The class should not expand `operator ==`'s parameter type from `Object`.

rrousselGit:
What's the downside to this exactly?

PR #1017, opened 2023-12-07
Keep operator== parameter type as Object

srawlins:
Fixes https://github.com/rrousselGit/freezed/issues/1016

### DIFF
--- a/packages/freezed/lib/src/templates/concrete_template.dart
+++ b/packages/freezed/lib/src/templates/concrete_template.dart
@@ -478,7 +478,7 @@ String toString($parameters) {
 
     return '''
 @override
-bool operator ==(dynamic other) {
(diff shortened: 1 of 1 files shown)


rrousselGit:
I appreciate the PR, but I don't fully understand the problem.

What are we _really_ solving here? Performance?

rrousselGit:
I was thinking that this could be useful in cases someone wants to override == by hand.
But it looks like from the interface POV, == isn't overridden, so it doesn't block them.


So this looks to be purely syntactic. Which I don't think adds a lot of value for generated code.

I'll merge it anyway, but if what triggered this PR is a lint rule, chances are that lint rule shouldn't trigger on generated files.
Users generally shouldn't read its source. Stylistic choices there shouldn't matter.

PR #1046, opened 2024-03-11
Fix the line ending symbol in JSON in this Doc comment

riscait:
Thanks again for the great package!

There was a minor error in the Doc comments that has been corrected.


rrousselGit:
Thanks!

PR #1073, opened 2024-04-18
build: add melos

altynbek132:
it generates pubspec overrides automatically which is useful for development
and pub get is faster with melos bootstrap


rrousselGit:
Thanks! But I don't want to use its more recent versions, because I don't want to introduce an extra configuration file in my projects.

PR #1081, opened 2024-05-03
analyzer <6.5.0 

SunlightBro:
analyzer 6.5.0 adds initial support for macros and has a bunch of deprecations.
So that the github-actions don't fail using flutter-master for now,
I added an upper-bound for package analyzer.


rrousselGit:
Thanks, but I prefer when the CI is failing.

That's a reminder that an action on my part is needed.

PR #1094, opened 2024-05-25
Update README

### LINKED ISSUE #1078: Confusing legacy tag for "Union types and Sealed classes"
## Context
The documentation marks the whole chapter of [Union types and Sealed classes](https://pub.dev/packages/freezed#legacy-union-types-and-sealed-classes) as Legacy. It gives an impression that this whole thing should be avoided and be better done in native dart. It gives a note explaining the details, and the case for `map/when` vs `switch` is clear and contains accurate reasoning. 

## Problem 
Nevertheless, if I am not mistaken, Union types and Sealed classes themselves are not outdated and are still relevant? It's just `map/when` should be replaced with `switch` and alike? 

Relevant use-case for union types: parsing json and convert it to a specific model by `unionKey`.

## Solution
Re-structure chapter in the way that block about `map/when` specifically is noted as legacy, but not union types as a whole.

## Additional context
As of Dart 3, Dart now has built-in pattern-matching using sealed classes.
As such, you no-longer need to rely on Freezed's generated methods for pattern matching. Instead of using when/map, use the official Dart syntax.

lishaduck:
I've updated the readme to resolve #1078, as it had confused me the first time I read through the documentation oh-so-long-ago.
The first commit contains the actual changes, the second commit is just formatting. It was a tiny diff, so I decided so PR it too, but separately for reviewability. Only the first commit makes actual changes to the content.
The only question I had was if there should be documentation about maybeMap and maybeWhen? As it's being phased out, I just deleted the broken reference. 
Thanks for the great tool! I use it all the time, and it makes life so much easier!


rrousselGit:
Thanks!

lishaduck:
> Thanks!

You're welcome!

PR #1095, opened 2024-05-26
Clarify `unfreezed` documention

lishaduck:
Just a minor copyedit. I was looking at the source to see if `@freezed` implied `@immutable` (which I assumed it did, but it doesn't), and noted that the docs were a bit unclear.


rrousselGit:
Thanks!

PR #1106, opened 2024-07-09
Bump json_serializable/json_annotation versions

### LINKED ISSUE #1105: Bump json_annotation and json_serializable versions
Bump both version to their most recent

* `json_annotation: ^4.9.0`
* `json_serializable: ^6.8.0`

The changelog for 2.5.5 - 2024-07-09 mentions the wrong package as well, should be `json_annotation: ^4.8.0`.

lrsvmb:
Closes #1105. 

Maybe modify the changelog as well?

Local tests passing.

### DIFF
--- a/benchmarks/pubspec.yaml
+++ b/benchmarks/pubspec.yaml
@@ -9,12 +9,12 @@ dependencies:
     sdk: flutter
   freezed_annotation:
     path: ../packages/freezed_annotation
-  json_annotation: ^4.4.0
+  json_annotation: ^4.9.0
 
 dev_dependencies:
   freezed:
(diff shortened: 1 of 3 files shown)


rrousselGit:
What's the purpose in such a change? Freezed doesn't rely on features from the latest version.

lrsvmb:
`json_serializable: ^6.7.0` allows for record types which I wanted to look into further, but you are right; that might not be necessary atm. Will close.

rrousselGit:
There's no public dependency on json_serializable. Only dev_dependency

So you can use any json_serializable version

PR #1113, opened 2024-07-24
Update README.md

Chematronix:
Explain what the private empty constructor does, include error so this shows up in search results

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -423,7 +423,7 @@ class Person with _$Person {
 
 then it will fail with the error `The non-abstract class _$_Person is missing implementations for these members:`.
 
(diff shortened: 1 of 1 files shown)


packages/freezed/README.md
```diff
@@ -421,9 +421,9 @@ class Person with _$Person {
 }
 ```
 
-then it won't work.
+then it will fail with the error `The non-abstract class _$_Person is missing implementations for these members:`.
 
-For that to work, we need an extra step: We need to define a private empty constructor:
+For that to work, we need to define a private empty constructor, which will signal the generated code to *extend/subclass* our class, instead of *implementing* it (which is the default, and only inherits type, and not properties or methods):
```
rrousselGit, line 426:
```suggestion
For that to work, we need to define a private empty constructor. That will enable the generated code to *extend/subclass* our class, instead of *implementing* it (which is the default, and only inherits type, and not properties or methods):
```

rrousselGit:
Thanks!

PR #1121, opened 2024-08-31
feat: add finalize parameter to Freezed config

Yegair:
Adds a new configuration parameter `finalize: bool` that allows to mark generated classes as `final` in the case of concrete classes or `sealed` in the case of abstract classes. It is disabled by default.

Enabling it allows the Dart analyzer/compiler to report patterns that will never match the type of a freezed class at compile time.

For example:

```dart
@Freezed(finalize: true)
sealed class Foo with _$Foo {
  const Foo._();
  const factory Foo() = _Foo;
}

@Freezed(finalize: true)
sealed class Bar with _$Bar {
  const Bar._();
  const factory Bar() = Bar;
}

void main() {
  switch (Foo()) {
    // This will now cause an analyzer warning: pattern_never_matches_value_type.
    // Without finalize: true it would not cause any warning/error at all.
    case Bar():
      print("matched Bar()");
      return;

    case Foo():
      print("matched Foo()");
      return;
  }
}
```

# Why do I think this is a good idea?

I have recently had an issue in one of my projects where I have a `@freezed sealed class DataFromApi` and I introduced a new `@freezed sealed class DataFromDatabase` that was intended to be used in many but not all cases where `DataFromApi` was being used so far.

It was a normal change, but I messed up due to the heavy use of pattern matching like

```dart
switch (data) {
  case DataFromApi(someCondition: true):
    // ...
    break;

  case DataFromApi(someCondition: false):
    // ...
    break;

  default:
    // ...
    break;
}
```

The changes mostly looked like this

```diff
- final data = await loadDataFromApi();
+ final data = await loadDataFromDatabase();

switch (data) {
  case DataFromApi(someCondition: true):
    // ...
    break;

    // ...
}
```

I was expecting the Dart analyzer/compiler to yell at me when changing the type of `data` in the example above to `DataFromDatabase`, but that was not the case. So I had to find all the now broken patterns by hand. Naturally I overlooked one case and the tests didn't catch it, so I got a Bug in production 😅. 

IMHO the Dart analyzer/compiler should be able to help me in this case no matter if I use freezed or not, since the classes in question were marked as `sealed`. So I opened an issue over there: https://github.com/dart-lang/sdk/issues/56616

However, since this might never be implemented or maybe is completely impossible, I thought it should be possible and rather simple to make freezed help me out in that case. All that needs to be done is mark the generated classes as `final` or `sealed`, and then the analyzer will step in and report any pattern that can never match at compile time.

# What needs to be done before it might be merged?

Good question, I think first of all someone needs to decide if this even is a good idea or if it would mess things up. In case it should be merged, I assume a few more tests need to be written to make sure the `finalize` flag works well with other configuration options. However, so far I have a hard time figuring out what even might break due to this change, so I would need help to decide what special/edge cases to consider when writing tests.

### DIFF
--- a/packages/freezed/test/source_gen_src.dart
+++ b/packages/freezed/test/source_gen_src.dart
@@ -224,3 +224,30 @@ abstract class AstractClass {
 class _AstractClass implements AstractClass {
   const _AstractClass();
 }
+
+@ShouldThrow(
+  '@freezed classes configured with [finalize: true] must be sealed',
+)
+@Freezed(finalize: true)
+class FinalizedClassWithoutSealedKeyword {
+  FinalizedClassWithoutSealedKeyword._();
+  factory FinalizedClassWithoutSealedKeyword() =
+      _FinalizedClassWithoutSealedKeyword;
+}
+
+class _FinalizedClassWithoutSealedKeyword
+    implements FinalizedClassWithoutSealedKeyword {
+  _FinalizedClassWithoutSealedKeyword();
+}
+
+@ShouldThrow(
+  '@freezed classes configured with [finalize: true] require a MyClass._() constructor',
+)
+@Freezed(finalize: true)
+sealed class FinalizedClassWithoutPrivateConstructor {
+  factory FinalizedClassWithoutPrivateConstructor() =
+      _FinalizedClassWithoutPrivateConstructor;
+}
+
+class _FinalizedClassWithoutPrivateConstructor
+    implements FinalizedClassWithoutPrivateConstructor {}
--- a/packages/freezed_annotation/lib/freezed_annotation.dart
+++ b/packages/freezed_annotation/lib/freezed_annotation.dart
@@ -219,6 +219,7 @@ class Freezed {
     this.makeCollectionsUnmodifiable,
     this.addImplicitFinal = true,
     this.genericArgumentFactories = false,
+    this.finalize = false,
   });
 
   /// Decode the options from a build.yaml
@@ -475,6 +476,46 @@ class Freezed {
   /// }
   /// ```
   final bool genericArgumentFactories;
+
+  /// Whether to add `sealed`/`final` modifiers to the generated classes.
+  ///
+  /// Defaults to false.
+  ///
+  /// This makes the generated classes `sealed` and `final` by default,
+  /// so when using them in a switch statement, the analzyer will warn you
+  /// if you try to match agains a pattern that will never match the type.
+  ///
+  /// When configuring a class with `finalize: true`, it also needs to be
+  /// `sealed` and have a private no-args constructor.
+  ///
+  /// ```dart
+  /// @Freezed(finalize: true)
+  /// sealed class Foo with _$Foo {
+  ///   const Foo._();
+  ///   const factory Foo() = _Foo;
+  /// }
+  ///
+  /// @Freezed(finalize: true)
+  /// sealed class Bar with _$Bar {
+  ///   const Bar._();
+  ///   const factory Bar() = _Bar;
+  /// }
+  ///
+  /// void main() {
+  ///   switch (Foo()) {
+  ///     // The analyzer will yield a warning that this case can never match,
+  ///     // because all subclasses of Foo are sealed/final, so it is guaranteed
+  ///     // that instances of type Bar can never also be of type Foo.
+  ///     case Bar():
+  ///       // ...
+  ///       break;
+  ///
+  ///     case Foo():
+  ///       // ...
+  ///       break;
+  ///   }
+  /// ```
+  final bool finalize;
 
   /// Options for customizing the generation of `map` functions
   ///
--- a/packages/_internal/lib/models.dart
+++ b/packages/_internal/lib/models.dart
@@ -102,6 +102,7 @@ class Data with _$Data {
     required GenericsParameterTemplate genericsParameterTemplate,
     required bool shouldUseExtends,
     required bool genericArgumentFactories,
+    required bool shouldMarkSealedOrFinal,
   }) = _Data;
 }
 
--- a/packages/_internal/lib/models.freezed.dart
+++ b/packages/_internal/lib/models.freezed.dart
@@ -1098,6 +1098,7 @@ mixin _$Data {
       throw _privateConstructorUsedError;
   bool get shouldUseExtends => throw _privateConstructorUsedError;
   bool get genericArgumentFactories => throw _privateConstructorUsedError;
+  bool get shouldMarkSealedOrFinal => throw _privateConstructorUsedError;
 
   /// Create a copy of Data
   /// with the given fields replaced by the non-null parameter values.
@@ -1126,7 +1127,8 @@ abstract class $DataCopyWith<$Res> {
       GenericsDefinitionTemplate genericsDefinitionTemplate,
       GenericsParameterTemplate genericsParameterTemplate,
       bool shouldUseExtends,
-      bool genericArgumentFactories});
+      bool genericArgumentFactories,
+      bool shouldMarkSealedOrFinal});
 
   $MapConfigCopyWith<$Res> get map;
   $WhenConfigCopyWith<$Res> get when;
@@ -1163,6 +1165,7 @@ class _$DataCopyWithImpl<$Res, $Val extends Data>
     Object? genericsParameterTemplate = null,
     Object? shouldUseExtends = null,
     Object? genericArgumentFactories = null,
+    Object? shouldMarkSealedOrFinal = null,
   }) {
     return _then(_value.copyWith(
       name: null == name
@@ -1229,6 +1232,10 @@ class _$DataCopyWithImpl<$Res, $Val extends Data>
           ? _value.genericArgumentFactories
           : genericArgumentFactories // ignore: cast_nullable_to_non_nullable
               as bool,
+      shouldMarkSealedOrFinal: null == shouldMarkSealedOrFinal
+          ? _value.shouldMarkSealedOrFinal
+          : shouldMarkSealedOrFinal // ignore: cast_nullable_to_non_nullable
+              as bool,
     ) as $Val);
   }
 
@@ -1276,7 +1283,8 @@ abstract class _$$DataImplCopyWith<$Res> implements $DataCopyWith<$Res> {
       GenericsDefinitionTemplate genericsDefinitionTemplate,
       GenericsParameterTemplate genericsParameterTemplate,
       bool shouldUseExtends,
-      bool genericArgumentFactories});
+      bool genericArgumentFactories,
+      bool shouldMarkSealedOrFinal});
 
   @override
   $MapConfigCopyWith<$Res> get map;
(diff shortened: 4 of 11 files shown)


packages/freezed/test/source_gen_src.dart
```diff
@@ -224,3 +224,30 @@ abstract class AstractClass {
...
+  factory FinalizedClassWithoutSealedKeyword() =
+      _FinalizedClassWithoutSealedKeyword;
+}
+
+class _FinalizedClassWithoutSealedKeyword
+    implements FinalizedClassWithoutSealedKeyword {
+  _FinalizedClassWithoutSealedKeyword();
+}
+
+@ShouldThrow(
```
rrousselGit, line 243: Why this requirement?

packages/freezed_annotation/lib/freezed_annotation.dart
```diff
@@ -219,6 +219,7 @@ class Freezed {
     this.makeCollectionsUnmodifiable,
     this.addImplicitFinal = true,
     this.genericArgumentFactories = false,
+    this.finalize = false,
```
rrousselGit, line 222: We likely want to default to `null`, and pick up the default from `build.yaml`

packages/freezed/test/source_gen_src.dart
```diff
@@ -224,3 +224,30 @@ abstract class AstractClass {
 class _AstractClass implements AstractClass {
   const _AstractClass();
 }
+
+@ShouldThrow(
+  '@freezed classes configured with [finalize: true] must be sealed',
```
rrousselGit, line 229: Why is this necessary?

rrousselGit:
This is valuable.

Although I'm wondering whether we need a flag at all. Why not always make the generated class `final` if it helps?
Extending a Freezed class isn't supported anyway, as they only have `factory` constructors.

packages/freezed/test/source_gen_src.dart
Yegair: I expected that sealed classes must be extended rather than implemented, but it seems it doesn't matter, so it can be removed 👍

packages/freezed/test/source_gen_src.dart
Yegair: Not marking the base class `sealed` more or less defeats the purpose of the entire change, because then the Dart analyzer can not figure out whether a pattern is "unmatchable".

However, from a technical perspective there is no need to seal the base class, so I guess it makes sense to remove that restriction

packages/freezed_annotation/lib/freezed_annotation.dart
Yegair: Will do that, but I wonder if there is an easy way to write a test that makes sure the value from the `build.yaml` is picked up correctly. Or is that something that can be generally considered as working and one doesn't need to care about it when just adding one more parameter?

Yegair:
Happy to hear that 😊

If I would implement that change in the scope of my project I would certainly make `final`/`sealed` the default. However, I find it very hard to anticipate how freezed is being used out there in the wild and making it the default feels like doing a breaking change. So I would ask you to make that decision, but for me at least it would definitely work.

packages/freezed_annotation/lib/freezed_annotation.dart
rrousselGit: There's the integration folder for build.yaml stuff

Yegair: It is now nullable and gets its default value (`false`) from its `@JsonKey` annotation

packages/freezed/test/source_gen_src.dart
Yegair: Removed it

packages/freezed/test/source_gen_src.dart
Yegair: Removed it

packages/freezed/pubspec_overrides.yaml
```diff
@@ -13,3 +13,11 @@ dependency_overrides:
   quiver: ^3.2.0
   # watcher 1.0.0 extends, implements, or mixens sealed class 'FileSystemEvent'
   watcher: ^1.1.0
+  # There seems to be a regression with Flutter ^3.24.0
+  # that causes an error when generating sources using the build_runner
+  # after running `flutter pub downgrade` in GitHub actions,
+  # because it then installs frontend_server_client: ^3.0.0:
+  #  Could not find a command named
+  #  "/path/to/flutter/stable-3.24.0-x64/bin/cache/dart-sdk/bin/snapshots/frontend_server.dart.snapshot".
+  # see https://github.com/dart-lang/build/issues/3733#issuecomment-2272082820
```
rrousselGit, line 22: Thanks for this!

Yegair:
From my side the PR is now "ready", meaning I have no further actionable task on my list.

Would be happy to remove a whole bunch of code and just hardcode `sealed` and `final` when generating subclasses, but as I said before: I would 100% support this, but I don't think I am in a position to make that decision.

Tried to make the CI work, but I guess the reason it failed has nothing to do with my changes, so I just added a dependency override for `frontend_server_client: ^4.0.0`. However, I am unsure if this might have negative side effects, so would be happy to remove it if requested.

Also I was thinking if `finalize` is a good term to use (I'm not a native english speaker), but maybe `makeGeneratedClassesFinal` or something in this direction would be easier to understand for users?

And finally, I tried to add a changeset as the bot requested, but I was unable to make it work by following the guide it linked to, so I assume using this package is somewhat outdated?

rrousselGit:
> Would be happy to remove a whole bunch of code and just hardcode sealed and final when generating subclasses, but as I said before: I would 100% support this, but I don't think I am in a position to make that decision.

You can go ahead and do this change. Only `_$Foo` classes are sub-classable at the moment, and that's not something folks should be using anyway.

Removing the flag and making the generated classes `final` are beneficial to more people this way. Otherwise most folks won't even know that the option is there and won't benefit from the warning.

rrousselGit:
You can ignore the changeset bot btw

packages/freezed/pubspec_overrides.yaml
Yegair: You're welcome!

Yegair:
Removed the configuration part, so basically it now just adds the `final`/`sealed` modifiers and of course left in some tests that make sure it works as intended.

I think it should be ready to merge.

If there is something else that should be changed, just let me know, but throughout the week it might take a while until I can get to it.

---

And since I get the chance to talk to you directly, just wanted to let you know how valuable Freezed and especially Riverpod are to the project I am working on (https://choosy.de if you want to know what people are building with your libs 😉). Since migrating from Bloc to Riverpod the ease of using providers from inside other providers made such a huge difference.

Cant wait to get a first glimpse of the wizardry you (hopefully) are going to do with macros once they are stable 😊.

rrousselGit:
My appologies, I kind of forgot about this PR. Let me see what we can do here.
Feel free to ping me if I forget again.

Yegair:
Friendly Reminder to take a look at this PR 😊

If there is a way I can help testing it, just let me know!

Yegair:
> Is this PR stil relevant as now freezed support "extends" ?

At least the original way how it was implemented will no longer work. Just did a quick rebase from upstream, but have to write quite a few more tests to make sure it works in combination with the new features from release `3.x.x`.

Will take a closer look on the weekend!

rrousselGit:
Sorry as you can see, I got sidetracked.  
I'm not sure how relevant it is anymore. To be honest, looking back at the motivation, I wonder if the iDE isn't to blame instead.

I'm not sure whether changing the generated code makes sense, when analyzer likely should know already that the `case` doesn't make sense for a sealed class.

After-all, `sealed` classes are some form of `final` classes

rrousselGit:
Actually ignore me, I can see why the analyzer works this way.

if you don't mind fixing the conflicts, we can merge this.


For now, let's not enable it by default though. I'd rather not make a 4.0 right after a 3.0

Yegair:
Updated the implementation, should be ready to review

- added new parameter `@Freezed(makeGeneratedClassesFinal: true)` which defaults to `null`.
- if `makeGeneratedClassesFinal` is `null`, it should be treated as if it was `false`, to not introduce a breaking change
- added a few more tests, mostly to see if it works with custom super-/sub-classes

To really test it, I integrated it with the project I am working on (which contains hundreds of freezed classes). However, due to https://github.com/RevenueCat/purchases-flutter/issues/1288 I am currently stuck on version `2.4.x`. Had to add a dependency override for `freezed_annotation: 3.0.0`. If I'm not mistaken, it should not matter which version of `freezed_annotation` is present at runtime, since everything happens at build time. So my reasoning was that I can safely ignore any transitive `freezed_annotation` dependencies, because naturally those dependencies have already been built, but TBH I do still don't fully understand how `build_runner` works, so I am not entirely sure.

rrousselGit:
Thinking about it, I wonder if this should be slightly generalised. We could have `@Freezed(classKeywords: ['final'])`

This would support `base`/`mixin`/`sealed`/...

Yegair:
I do like the idea of using a more general API. Of course we can just use a `List<String>`, but it should also be possible to do something like

```Dart
@Freezed(classModifiers: [FreezedClassModifier.Final])
abstract class Foo with _$Foo {
  factory Foo() = _Foo;
}
```

Might be a little more developer friendly, because one would get autocomplete. However, due to all modifiers being keywords we either have to use UpperCamelCase (like in the example) or something like `FreezedClassModifier.$final`.

When it comes to which modifiers should be supported, I think everything that makes the class effectively abstract is problematic, for example:

```Dart
@Freezed(classModifiers: [FreezedClassModifier.Sealed])
abstract class Foo with _$Foo {
  factory Foo() = _Foo;
               // ^^^^
               // The redirecting constructor 'Foo' can't redirect to a constructor of the abstract class '_Foo'.
}
```

So that would leave the modifiers `final`, `mixin`, `base`, unless there is a way of using `abstract`, `sealed`, `interface` on the generated classes that I am not aware of.

---

That being said, I do have one "concern" with that change. Allowing more modifiers might lead to people using Freezed in unintended ways, which in turn might make future features more difficult to implement without breaking stuff. However, that is just a gut feeling that I have, and I can not support it with examples.

If I would have to make the decision, I would probably start by allowing `final` and `base` only. Then I'd wait and see if anyone asks for more modifiers.

Yegair:
Did a quick experiment on it. Let me know what you think of it. Not sure why CI is currently failing (something with Dart Format 🤔), tests should still be passing though.

rrousselGit:
Ah, I kind of forgot that we wouldn't be able to mark the generated class `sealed`. Even though Freezed has inheritance now, the factory constructor still wouldn't work.

So we'd only gain `mixin/base`. That's a bit limited, so probably not worth the trouble

Do you mind reverting your last commit? I'll merge this then. I've already reviewed the previous changes.

Yegair:
@rrousselGit Just rebased this one onto the latest `master`. Do you think it could be merged? Would love to get rid of my forked freezed 😅

PR #1189, opened 2025-03-01
Add migration guide

### LINKED ISSUE #1187: Add migration guide to v3
The v3 release introduced many breaking changes. A migration guide would help developer to quickly adapt to these new changes.

Bonus: For trivial changes like adding the "abstract" keyword a dart fix could be written.

snapsl:
closes #1187 

## Summary by CodeRabbit

- **Documentation**
	- Added a new "Migration" section to provide users with guidance on transitioning between package versions.
	- Updated the migration guide to highlight required syntax changes and revised usage practices, assisting users in adapting to the latest version.


snapsl:
I just recognized that you can generate unions without making them sealed.

```dart
@freezed
class Union with _$Union {
  const factory Union.data(int value) = Data;
  const factory Union.loading() = Loading;
  const factory Union.error([String? message]) = Error;
}
```
Is there any use case to this?

rrousselGit:
> Is there any use case to this?

There's no reason to require the `sealed` if it works without. Folks do what they want

snapsl:
@rrousselGit  Thanks for clarification.
If there are no benefits, I would require a sealed here to mitigate any potential problems.

rrousselGit:
Classes can be concrete. I cannot require abstract/sealed.

packages/freezed/README.md
```diff
@@ -1247,6 +1248,10 @@ targets:
             when_or_null: false
 ```
 
+# Migration
```
rrousselGit, line 1251: This should be at the top, not the bottom

PR #1222, opened 2025-04-04
Fix  (#1223) Classic class copyWith Adds Unnecessary Bang Operator Nullable Unspecified Generic Types, Causing Runtime Exceptions

### LINKED ISSUE #1223: copyWith Adds Unnecessary Bang Operator to Classic Style Nullable Generic Unspecified Types, Causing Runtime Exceptions
### Describe the bug
Classic style classes with nullable generics where the generic is unspecified generate a `copyWith` method that throws a `_TypeError (Null check operator used on a null value)` when called with no arguments (i.e., `.copyWith()`) if the current value of a nullable field is `null`. This occurs because Freezed adds a bang operator (`!`) to the field access, even when the field and constructor parameter are both nullable.

### Example:

```dart
@freezed
abstract class Inner<I> with _$Inner<I> {
  const factory Inner({I? data}) = _Inner<I>;
}

@freezed
class ClassicUnspecifiedOuter with _$ClassicUnspecifiedOuter {
  @override
  // ignore: strict_raw_type
  final Inner? innerData;

  const ClassicUnspecifiedOuter({this.innerData});
}
```

Generated copyWith:
```
class _$ClassicUnspecifiedOuterCopyWithImpl<$Res>
    implements $ClassicUnspecifiedOuterCopyWith<$Res> {
  _$ClassicUnspecifiedOuterCopyWithImpl(this._self, this._then);

  final ClassicUnspecifiedOuter _self;
  final $Res Function(ClassicUnspecifiedOuter) _then;

  /// Create a copy of ClassicUnspecifiedOuter
  /// with the given fields replaced by the non-null parameter values.
  @pragma('vm:prefer-inline')
  @override
  $Res call({
    Object? innerData = freezed,
  }) {
    return _then(ClassicUnspecifiedOuter(
      innerData: freezed == innerData
          ? _self.innerData!
          : innerData // ignore: cast_nullable_to_non_nullable
              as Inner<dynamic>?,
    ));
  }
}
```

The culprit is:

```
innerData: freezed == innerData
          ? _self.innerData!
          : innerData // ignore: cast_nullable_to_non_nullable
              as Inner<dynamic>?,
```

This throws an exception if `_self.innerData` is null and no value is passed to copyWith.

### To Reproduce

1. Define the class hierarchy above.
2. Run dart run build_runner build to generate the Freezed code.
3. Create an instance: ClassicUnspecifiedOuter(innerData: null).
4. Call .copyWith()—it throws _TypeError (Null check operator used on a null value).

### Expected Behavior
No bang operator should be added when the field and constructor parameter are both nullable:

`innerData: freezed == innerData
          ? _self.innerData
          : innerData // ignore: cast_nullable_to_non_nullable
              as Inner<dynamic>?,`

### Actual Behavior
A bang operator is added, causing runtime exceptions:

`innerData: freezed == innerData
          ? _self.innerData!
          : innerData // ignore: cast_nullable_to_non_nullable
              as Inner<dynamic>?,`

**_BTW, if the inner generic does not extend anything, the generated code doesn't specify the `<dynamic>` and it creates a lint warning for strict_raw_type in the generated file, which is why i added `<I extends Object>` to get the build passing._**

BenjiFarquhar:
Fixes #1223 

Please see issue for full details.

I have done it in two commits, so you can check out the first commit to see the failing tests with original code, then check out the second commit and run the generator, then run the tests to see the passing tests. 

**_BTW, if the inner generic does not extend anything, the generated code doesn't specify the `<dynamic>` and it creates a lint warning for strict_raw_type in the generated file, which is why i added `<I extends Object>` to get the build passing._**

## Summary by CodeRabbit

- **New Features**
  - Introduced a new abstract class, `InnerExtends`, and a new immutable data class, `ClassicUnspecifiedOuter`, enhancing handling of optional inner values.

- **Refactor**
  - Improved the internal logic for managing type conversions and null-safety, ensuring more accurate behavior.

- **Tests**
  - Added test cases to verify that the `copyWith` method correctly handles scenarios with null inner data without triggering errors.

### DIFF
--- a/packages/freezed/lib/src/templates/copy_with.dart
+++ b/packages/freezed/lib/src/templates/copy_with.dart
@@ -286,7 +286,31 @@ $s''';
       }
 
       var cast = '';
-      if (propertyGetterForCopyWithParameter.type != to.type) cast = '!';
+      final fieldType = propertyGetterForCopyWithParameter.type;
+      final paramType = to.type;
+      final fieldIsNullable = propertyGetterForCopyWithParameter.isNullable;
+      final paramIsNullable = to.isNullable;
+
+      if (paramType == null) {
+        if (!fieldIsNullable) cast = '!';
+      } else {
+        final fieldBaseType = fieldType.endsWith('?')
+            ? fieldType.substring(0, fieldType.length - 1)
+            : fieldType;
+        final paramBaseType = paramType.endsWith('?')
+            ? paramType.substring(0, paramType.length - 1)
+            : paramType;
+
+        if (fieldIsNullable && !paramIsNullable) {
+          cast = '!';
+        } else if (!fieldIsNullable && paramIsNullable) {
+          cast = '';
+        } else if (!fieldIsNullable &&
+            !paramIsNullable &&
+            fieldBaseType != paramBaseType) {
+          cast = '!';
+        }
+      }
 
       return '$accessor.$propertyName$cast';
     }
--- a/packages/freezed/test/generic_test.dart
+++ b/packages/freezed/test/generic_test.dart
@@ -130,4 +130,11 @@ void main() {
       throwsCompileError,
     );
   });
+
+  test(
+      'copyWith null innerData does not throw when using classic class style, unspecified generic',
+      () {
+    final state = ClassicUnspecifiedOuter(innerData: null);
+    expect(() => state.copyWith(), returnsNormally);
+  });
 }
--- a/packages/freezed/test/integration/generic.dart
+++ b/packages/freezed/test/integration/generic.dart
@@ -66,3 +66,17 @@ abstract class Inner<I> with _$Inner<I> {
 abstract class Outer with _$Outer {
   const factory Outer({Inner<int>? innerData}) = _Outer;
 }
+
+@freezed
+abstract class InnerExtends<I extends Object> with _$InnerExtends<I> {
+  const factory InnerExtends({I? data}) = _InnerExtends<I>;
+}
+
+@freezed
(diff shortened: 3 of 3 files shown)


BenjiFarquhar:
Hi, there seems to be some failing tests, and it is quite hard to figure out if the test has mistake or the generator should handle that configuration. 

Such as this class:

```
@freezed
class ManualWithBothDefault with _$ManualWithBothDefault {
  ManualWithBothDefault._(String this.b) : a = null;
  ManualWithBothDefault(int this.a) : b = null;

  @override
  final int? a;
  @override
  final String? b;
}
```

I don't think that is valid dart, it is specifying `this` field b in the constructor with a different type than what the field is. 

And in these ones:

```
@freezed
abstract class Union with _$Union {
  factory Union.foo({int? arg}) = _UnionFoo;
  factory Union.bar({required int arg}) = _UnionBar;
}

```
I'm not sure if freezed handles the same field name twice, with different types. 

If it is supposed to handle that, then we may have a **very edge case** working when a **very normal case** doesn't work in stable currently. I would prefer to use this PR than stable because of that. 

Otherwise we just need to update the tests to align the types which is easy.

rrousselGit:
I appreciate, but the fix is a bit more involved than this. We typically don't want to rely on the type's toString to do type checks like `?` and such. I've handled this in a different PR

BenjiFarquhar:
> I appreciate, but the fix is a bit more involved than this. We typically don't want to rely on the type's toString to do type checks like `?` and such. I've handled this in a different PR

No problem at all. Just happy it is resolved. Thanks for the speedy fix!

PR #1226, opened 2025-04-05
Fix diagnostics

### LINKED ISSUE #1184: Freezed v3 doesn't generate diagnosticable properties if you override toString
> I'm updating the docs for #1181, and I found [`packages/freezed/example/lib/diagnosticable.dart`](https://github.com/rrousselGit/freezed/blob/master/packages/freezed/example/lib/diagnosticable.dart) is also impacted by this. It looks like if you override toString, Freezed won't generate the rest of the diagnosticable properties either. See #1183.
> 
> @rrousselGit, does that mean that the example not tested in CI? Would you want a PR for that? (Started on that in #1185)

 _Originally posted by @lishaduck in [#1182](https://github.com/rrousselGit/freezed/issues/1182#issuecomment-2691830109)_

### LINKED ISSUE #1200: Code generation changes with the libraries imported
**Describe the bug**

When adding the import of the foundation library, the code generated is not the same. The generated mixin implements `DiagnosticableTreeMixin` causing errors when using the classic class syntax.

**To Reproduce**

```dart
import 'package:flutter/foundation.dart';

// required: associates our `main.dart` with the code generated by Freezed
part 'main.freezed.dart';

@freezed
@JsonSerializable()
class Person with _$Person {
  const Person({
    required this.firstName,
    required this.lastName,
    required this.age,
  });

  final String firstName;
  final String lastName;
  final int age;
}
```

**Expected behavior**
I was expecting the code generated to not depend on whether or not I import the foundation library

rrousselGit:
fixes #1184
fixes #1200

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved diagnostic output generation to ensure consistency and correctness.
  - Simplified the diagnostic method signature to provide a more streamlined experience.
  - Refined the conditions under which diagnostic features are applied for enhanced reliability.

- **Chores**
  - Updated build configurations to extend testing coverage and optimize the workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,12 +17,11 @@ jobs:
       matrix:
         package:
           - packages/freezed
+          - packages/freezed/example
           - packages/freezed_annotation
           - packages/freezed_lint
         channel:
           - master
-          # BLocked by https://github.com/dart-lang/source_gen/issues/739
-          # - stable
         dependencies:
           - get
           - downgrade
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
       - uses: subosito/flutter-action@v2
         with:
           channel: ${{ matrix.channel }}
-          cache: ${{ matrix.channel == 'stable' }}
+          cache: ${{ matrix.channel == 'master' }}
 
       # It is executed separately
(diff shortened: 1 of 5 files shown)


lishaduck:
Ah, that works as well. Hadn't considered turning off diagnostics when overriding toString.

I might document that 🤔

PR #1268, opened 2025-06-30
docs: simplify dependency installation commands by grouping packages

EchoEllet:
Allows users to copy-paste commands and run them at once instead of copying each command and running it separately. This is also a bit more readable since `build_runner` is always required when using `freezed`, and both `freezed` and `freezed_annotation` are required at the same time to use Freezed. While `json_serializable` can be slightly unrelated since serialization is not part of data classes, and it makes sense to separate them in case the user doesn't need JSON serialization.

## Summary by CodeRabbit

* **Documentation**
  * Simplified installation instructions by combining related package dependencies into single command lines for Flutter and Dart projects.

### DIFF
--- a/packages/freezed/README.md
+++ b/packages/freezed/README.md
@@ -83,7 +83,10 @@ First, install [build_runner] and [Freezed] by adding them to your `pubspec.yaml
 For a Flutter project:
 
 ```console
-flutter pub add dev:build_runner freezed_annotation dev:freezed
+flutter pub add \
+  dev:build_runner \
+  freezed_annotation \
+  dev:freezed
 # if using freezed to generate fromJson/toJson, also add:
 flutter pub add json_annotation dev:json_serializable
 ```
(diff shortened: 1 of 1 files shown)


packages/freezed/README.md
```diff
@@ -83,23 +83,17 @@ First, install [build_runner] and [Freezed] by adding them to your `pubspec.yaml
 For a Flutter project:
 
 ```console
-flutter pub add freezed_annotation
-flutter pub add dev:build_runner
-flutter pub add dev:freezed
+flutter pub add dev:build_runner freezed_annotation dev:freezed
```
rrousselGit, line 86: We lose quite a bit in readability.

Could you use newlines to have one package per line?

EchoEllet: Is this better?

```dart
flutter pub add dev:build_runner
flutter pub add dev:build_runner freezed_annotation dev:freezed
```

rrousselGit: No I mean using a `\n` between each package name.

EchoEllet: So, is this what you expect?:

```dart
flutter pub add \
  dev:build_runner \
  freezed_annotation \
  dev:freezed
```

rrousselGit: Yes

rrousselGit:
Thanks!

PR #1279, opened 2025-07-18
fix: freezed_missing_mixin on private types due to leading underscore

refi64:
Private types get their leading underscore stripped off, but the lint wasn't accounting for that, so it would end up trying to change the type name to one that does not actually exist.


refi64:
@rrousselGit oops I didn't realize there *were* tests because I had badly misread how they work, should be added now.

packages/freezed_lint/example/lib/missing_mixin.dart
```diff
@@ -37,3 +37,16 @@ abstract class BarModel with _$BarModel, SomeMixin {
   @override
   int get id => id;
 }
+
+@freezed
+class _BazModel with _$BazModel {
+  const factory _BazModel() = __BazModel;
+}
+
+// fake generated code:
```
rrousselGit, line 46: Why are we using fake generated code?

refi64: hmm I might have misunderstood this setup again. the `.g.dart` file here can't be regenerated because of a hard error:

```
$ dart run build_runner build
1s freezed on 1 input: 1 no-op

Failed to build with build_runner in 2s; wrote 0 outputs.

log output for freezed on lib/missing_mixin.dart
E Classes using @freezed must use `with _$MissingMixin`.
  package:example_app/missing_mixin.dart:7:7
    ╷
  7 │ class MissingMixin {
    │       ^^^^^^^^^^^^
    ╵
  package:example_app/missing_mixin.dart:5:1
    ╷
  5 │ ┌ @freezed
  6 │ │ // expect_lint: freezed_missing_mixin
  7 │ │ class MissingMixin {
  8 │ │   const factory MissingMixin() = _MissingMixin;
  9 │ └ }
```

In retrospect I think this is actually accidental: the [change to make this an error in the source generator](https://github.com/rrousselGit/freezed/commit/71cca2e5f4bc45a6ce4b7716663d497ed0024217) is only a few months old (2025), whereas the last change to the `.freezed.dart` file here was in 2024. I'm not sure what other ideal way to get around this would be.

refi64: oh well that feels obvious in retrospect :sweat_smile: fixed

PR #1344, opened 2026-04-12
Allow analyzer 11.x and 12.x releases

ThexXTURBOXx:
Allow analyzer 11.x and 12.x releases

## Summary by CodeRabbit

* **Chores**
  * Updated analyzer dependency constraints in freezed and freezed_lint packages to support a broader range of compatible versions.

### DIFF
--- a/packages/freezed/pubspec.yaml
+++ b/packages/freezed/pubspec.yaml
@@ -10,7 +10,7 @@ environment:
   sdk: ">=3.8.0 <4.0.0"
 
 dependencies:
-  analyzer: '>=9.0.0 <11.0.0'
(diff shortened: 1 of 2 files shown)


ThexXTURBOXx:
The issues pointed out by the AI were already there before. They should not matter really.
After my PR is merged, older releases of analyzer will still (additionally) be compatible.

rrousselGit:
There are many breaking changes involved. This is not enough to upgrade. I'm dealing with this in a separate PR

ThexXTURBOXx:
Weird - when I tested it, I applied numerous dependency overrides such that I could at least run the dart analyzer on the code and it gave no errors...
But thanks for dealing with this then!

